PR #12, opened 2016-05-19
adds basic testing

gregcaporaso:
this is needed to confirm that some lower level changes that I'm working on aren't breaking things. 

@ebolyen, could you review this one?

### DIFF
--- a/q2_diversity/__init__.py
+++ b/q2_diversity/__init__.py
@@ -6,8 +6,8 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-__version__ = "0.0.0-dev"
-
 from ._beta import beta_diversity
+
+__version__ = "0.0.0-dev"
 
 __all__ = ['beta_diversity']
--- a/q2_diversity/tests/test_beta.py
+++ b/q2_diversity/tests/test_beta.py
@@ -5,3 +5,69 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
+
+import unittest
+import io
+
+import skbio
+import numpy as np
+import numpy.testing as npt
+from biom.table import Table
+
+from q2_diversity import beta_diversity
+
+
+class BetaDiversityTests(unittest.TestCase):
+
+    def test_non_phylogenetic(self):
+        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
+                  ['O1', 'O2'],
+                  ['S1', 'S2', 'S3'])
+        actual = beta_diversity('braycurtis', t)
+        # expected computed with scipy.spatial.distance.braycurtis
+        expected = skbio.DistanceMatrix([[0.0000000, 0.3333333, 0.6666667],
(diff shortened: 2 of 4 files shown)


q2_diversity/__init__.py
```diff
@@ -6,8 +6,8 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-__version__ = "0.0.0-dev"
-
 from ._beta import beta_diversity
 
+__version__ = "0.0.0-dev"
```
ebolyen: pip doesn't handle `0.0.0` well nor dashes, change to `0.0.1dev`

q2_diversity/tests/test_beta.py
```diff
@@ -5,3 +5,68 @@
...
+            for id2 in actual.ids:
+                npt.assert_almost_equal(actual[id1, id2], expected[id1, id2])
+
+    def test_non_phylogenetic_invalid_input(self):
+        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
+                  ['O1', 'O2'],
+                  ['S1', 'S2', 'S3'])
+        tree = skbio.TreeNode.read(io.StringIO(
+            '((O1:0.25, O2:0.50):0.25, O3:0.75)root;'))
+        self.assertRaises(TypeError, beta_diversity, 'braycurtis', t,
```
ebolyen:
```python
with self.assertRaises(TypeError):
    beta_diversity('bray_curtis', t, phylogeny=tree)
```

q2_diversity/tests/test_beta.py
```diff
@@ -5,3 +5,68 @@
...
+                          phylogeny=tree)
+
+    def test_phylogenetic_invalid_input(self):
+        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
+                  ['O1', 'O2'],
+                  ['S1', 'S2', 'S3'])
+        self.assertRaises(TypeError, beta_diversity, 'unweighted_unifrac', t)
```
ebolyen: Use `with` like above.

q2_diversity/tests/test_beta.py
```diff
@@ -5,3 +5,68 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
+
+from unittest import TestCase, main
```
ebolyen: We tend to just import `unittest`, but it's not that important.

ebolyen:
@gregcaporaso, done reviewing, only a few comments.

q2_diversity/__init__.py
gregcaporaso: This hasn't changed (just moved location in this file). Note this is what we're doing in the other plugins right now:
https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/__init__.py#L15
https://github.com/qiime2/q2-types/blob/master/q2_types/__init__.py#L12

Mind if we just leave it that way for the moment, and change it in all the repos at the same time?

ebolyen: sounds fine

ebolyen:
Thanks @gregcaporaso!

PR #39, opened 2016-10-04
ENH: adds bioenv method, fixes #20

### LINKED ISSUE #20: bioenv workflow
This would be nice for alpha, but it's not essential. It is easy though, as it's implemented in scikit-bio.

gregcaporaso:
Here's an example of the output:

screenshot 2016-10-03 20 32 36


jairideout:
👍 🎉 👍

PR #43, opened 2016-10-06
ENH/REF: Move beta into subpackage, brand a-corr.

thermokarst:


q2_diversity/_beta/_method.py
```diff
@@ -0,0 +1,62 @@
...
+    sample_ids = table.ids(axis='sample')
+
+    return skbio.diversity.beta_diversity(
+        metric=metric,
+        counts=counts,
+        ids=sample_ids
+    )
+
+
+_beta_group_significance_fns = {'permanova': skbio.stats.distance.permanova,
```
gregcaporaso: This should go to `q2_diversity/_beta/_visualizer.py`.

q2_diversity/_alpha/_visualizer.py
```diff
@@ -146,28 +151,28 @@ def alpha_correlation(output_dir: str,
...
-        fh.write(' <tr><td>Sample size</td><td>%d</td></tr>\n' %
-                 df.shape[0])
-        fh.write('</table>')
-        fh.write('<p>\n')
-        fh.write('<a href="scatter-plot.pdf">\n')
-        fh.write(' <img src="scatter-plot.png">')
-        fh.write(' <p>Download as PDF</p>\n')
-        fh.write('</a>\n\n')
-        fh.write('</body></html>')
+    warning = None
```
gregcaporaso: You've tested with and without a warning, right? Just want to confirm because I notice that the data type differs when there are/aren't warnings. The unit tests would catch this as it covers both cases.

q2_diversity/plugin_setup.py
```diff
@@ -99,7 +99,7 @@
 )
 
 beta_group_significance_methods = \
-    list(q2_diversity._beta._beta_group_significance_fns)
+    list(q2_diversity._beta._method._beta_group_significance_fns)
```
gregcaporaso: Be sure to update this when you move `_beta_group_significance_fns`.

PR #55, opened 2016-10-22
ENH: Upgrade visualizer rendering to use q2templates

jakereps:
Convert existing templates to use `q2templates`, and add templating for `bioenv` and `beta_group_significance`


thermokarst:
Looks like there are some failing tests here, mind updating them to pass with the new templates? Thanks @jakereps!

PR #62, opened 2016-11-02
ENH: Fix alpha-correlation minX and Y axis collisions for alpha-correlation

### LINKED ISSUE #45: alpha-correlation improvements
- [ ] set min x value for the x-axis to be the minimum x value in the data set minus some pad (e.g., 3% of x range) so that points don't show up on the y-axis - here's an example of the issue:

screenshot 2016-10-07 08 12 40

jakereps:
Fixes #45


q2_diversity/_alpha/alpha_correlation_assets/src/render.js
```diff
@@ -17,7 +17,8 @@ export function render(svg, data) {
 
   const { xAxisLabel, yAxisLabel, minX, maxX, minY, maxY } = data;
 
-  const x = scaleLinear().domain([minX, maxX]).range([0, width]).nice();
+  // Add a ~3% padding to minX values, also handle for a minX of 0
+  const x = scaleLinear().domain([minX * 0.97 || minX - 0.03, maxX]).range([0, width]).nice();
```
thermokarst: The `minX - 0.03` part isn't actually a percentage, so this produces some weird results for some metadata categories. How about:

```javascript
const x = scaleLinear().domain([minX - ((maxX - minX) * 0.03), maxX]).range([0, width]).nice();
```

q2_diversity/_alpha/alpha_correlation_assets/package.json
```diff
@@ -20,8 +20,8 @@
     "webpack-dev-server": "^1.16.1"
   },
   "scripts": {
-    "dev": "./node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline",
-    "build": "./node_modules/webpack/bin/webpack.js"
+    "dev": "webpack-dev-server --inline",
```
thermokarst: 👍

q2_diversity/_alpha/alpha_group_significance_assets/package.json
```diff
@@ -20,8 +20,8 @@
     "webpack-dev-server": "^1.16.1"
   },
   "scripts": {
-    "dev": "./node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline",
-    "build": "./node_modules/webpack/bin/webpack.js"
+    "dev": "webpack-dev-server --inline",
```
thermokarst: 👍

jakereps:
Thanks @thermokarst, should be good to go after tests 👍 ! The scaling was kind of odd, as I'd set the min to -0.03 (on that initial commit) and it'd stretch it down to -2 as the starting point.

q2_diversity/_alpha/alpha_correlation_assets/src/render.js
jakereps: That one was just to make it non-zero after it fell back on the `||` for `minX` being 0.

PR #72, opened 2016-12-09
ENH: Filter zero-variance numerical columns from bioenv

### LINKED ISSUE #56: bioenv fails on metadata with 0 variance
See [this forum discussion](https://forum.qiime2.org/t/diversity-bioenv-column-s-in-the-data-frame-could-not-be-scaled/58/2) to see an example error and more details.

If the metadata map contains any columns with 0 variance, `bioenv` fails.

@gregcaporaso  suggests: 
`What we should probably do is detect these columns, not include them in the analysis, and print a warning in the resulting visualization about columns that were excluded, similar to what we do in the other visualizations in this plugin`

It could be useful to specify categories as an optional parameter to `bioenv`, e.g., to reduce runtime on very large metadata maps (such as from `qiita`) and/or to avoid certain categories.

jakereps:
Resolves #56 

As this sits it just filters and alerts the user of what was removed. Should specifying optional categories be added to the functionality?

image

### DIFF
--- a/q2_diversity/tests/test_beta.py
+++ b/q2_diversity/tests/test_beta.py
@@ -111,7 +111,7 @@ class BioenvTests(unittest.TestCase):
             index_fp = os.path.join(output_dir, 'index.html')
             self.assertTrue(os.path.exists(index_fp))
             self.assertTrue('metadata1' in open(index_fp).read())
-            self.assertFalse('metadata2' in open(index_fp).read())
+            self.assertTrue('metadata2' in open(index_fp).read())
             self.assertFalse('Warning' in open(index_fp).read())
 
     def test_bioenv_exclude_missing_data(self):
@@ -148,7 +148,7 @@ class BioenvTests(unittest.TestCase):
             index_fp = os.path.join(output_dir, 'index.html')
             self.assertTrue(os.path.exists(index_fp))
             self.assertTrue('metadata1' in open(index_fp).read())
-            self.assertFalse('metadata2' in open(index_fp).read())
+            self.assertTrue('metadata2' in open(index_fp).read())
             self.assertFalse('Warning' in open(index_fp).read())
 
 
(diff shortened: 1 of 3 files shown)


jairideout:
> As this sits it just filters and alerts the user of what was removed. Should specifying optional categories be added to the functionality?

That sounds useful but probably worth adding that functionality in a separate PR. I don't think this is high priority for the upcoming workshop. Can you please create an issue so we can keep track of it?

jairideout:
Can you format the message so that the `dict` repr isn't displayed? For example, this is how the other plugins format the message:

https://forum-qiime2-org.s3-us-west-2.amazonaws.com/original/1X/6943c7495bcf8f9f4e8f1f30eacfaef096cbc783.png

q2_diversity/tests/test_beta.py
```diff
@@ -111,7 +111,7 @@ def test_bioenv(self):
             index_fp = os.path.join(output_dir, 'index.html')
             self.assertTrue(os.path.exists(index_fp))
             self.assertTrue('metadata1' in open(index_fp).read())
-            self.assertFalse('metadata2' in open(index_fp).read())
+            self.assertTrue('metadata2' in open(index_fp).read())
```
jairideout: Can you update these tests to be a little more specific? i.e. that `metadata2` column is in the "non-numeric" warning?

Can you also add a unit test that tests out the zero-variance functionality?

q2_diversity/tests/test_beta.py
```diff
@@ -148,7 +148,7 @@ def test_bioenv_extra_metadata(self):
...
```
jairideout: Same comment here

jakereps:
Thanks @jairideout! Clarified the test assertions a bit, and added a no variance test as well!

jairideout:
Looks great, thanks @jakereps!! 👍 💯 👍

PR #74, opened 2016-12-14
ENH: Rephrase skbio warnings referring to OTUs

### LINKED ISSUE #59: error message coming from skbio in `alpha/beta_phylogenetic` refers to "otu ids"
We should update this to reference feature ids for clarity:

```
MissingNodeError: All ``otu_ids`` must be present as tip names in ``tree``.
```

jakereps:
Resolves #59

### DIFF
--- a/q2_diversity/_alpha/_method.py
+++ b/q2_diversity/_alpha/_method.py
@@ -22,7 +22,7 @@ def non_phylogenetic_metrics():
             'dominance', 'doubles', 'enspie', 'esty_ci', 'fisher_alpha',
             'goods_coverage', 'heip_e', 'kempton_taylor_q', 'margalef',
             'mcintosh_d', 'mcintosh_e', 'menhinick', 'michaelis_menten_fit',
-            'feature_counts', 'osd', 'pielou_e', 'robbins', 'shannon',
+            'observed_otus', 'osd', 'pielou_e', 'robbins', 'shannon',
             'simpson', 'simpson_e', 'singles', 'strong', 'gini_index',
             'lladser_pe', 'lladser_ci'}
 
@@ -43,12 +43,9 @@ def alpha_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
                                                  otu_ids=feature_ids,
                                                  tree=phylogeny)
     except skbio.tree.MissingNodeError as e:
-        if 'otu_ids' in str(e):
-            raise skbio.tree.MissingNodeError("All ``feature_ids`` must be "
-                                              "present as tip names in "
-                                              "``tree``.")
-        else:
-            raise
+        message = str(e).replace('otu_ids', 'feature_ids')
+        message = message.replace('tree', 'phylogeny')
+        raise skbio.tree.MissingNodeError(message)
 
     result.name = metric
     return result
@@ -58,12 +55,10 @@ def alpha(table: biom.Table, metric: str) -> pd.Series:
     if metric not in non_phylogenetic_metrics():
         raise ValueError("Unknown metric: %s" % metric)
 
-    new_metric = metric if metric != 'feature_counts' else 'observed_otus'
-
     counts = table.matrix_data.toarray().astype(int).T
     sample_ids = table.ids(axis='sample')
 
-    result = skbio.diversity.alpha_diversity(metric=new_metric, counts=counts,
+    result = skbio.diversity.alpha_diversity(metric=metric, counts=counts,
                                              ids=sample_ids)
     result.name = metric
     return result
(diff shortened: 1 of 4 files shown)


q2_diversity/_alpha/_method.py
```diff
@@ -36,12 +36,15 @@ def alpha_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
-                otu_ids=feature_ids,
-                tree=phylogeny)
+    try:
+        result = skbio.diversity.alpha_diversity(metric=metric,
+                                                 counts=counts,
+                                                 ids=sample_ids,
+                                                 otu_ids=feature_ids,
+                                                 tree=phylogeny)
+    except skbio.tree.MissingNodeError:
+        raise skbio.tree.MissingNodeError("All ``feature_ids`` must be "
```
thermokarst: From what I can tell we can possibly have a few variants of the `MissingNodeError` raised here (@jairideout correct me if I am mistaken). If this is the case, we probably want to check the error message itself and make sure we are replacing the right one.

q2_diversity/_beta/_method.py
```diff
@@ -35,13 +36,17 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
-        otu_ids=feature_ids,
-        tree=phylogeny
-    )
+    try:
+        results = skbio.diversity.beta_diversity(metric=metric,
+                                                 counts=counts,
+                                                 ids=sample_ids,
+                                                 otu_ids=feature_ids,
+                                                 tree=phylogeny)
+    except skbio.tree.MissingNodeError:
```
thermokarst: Same as above.

jakereps:
If there are other `MissingNode` errors they should pass right through now. Thanks @thermokarst!

q2_diversity/_alpha/_method.py
jairideout: Yeah I think there are a couple different ways this error type can be raised.

q2_diversity/_alpha/_method.py
```diff
@@ -22,7 +22,7 @@ def non_phylogenetic_metrics():
             'dominance', 'doubles', 'enspie', 'esty_ci', 'fisher_alpha',
             'goods_coverage', 'heip_e', 'kempton_taylor_q', 'margalef',
             'mcintosh_d', 'mcintosh_e', 'menhinick', 'michaelis_menten_fit',
-            'observed_otus', 'osd', 'pielou_e', 'robbins', 'shannon',
+            'feature_counts', 'osd', 'pielou_e', 'robbins', 'shannon',
```
jairideout: @gregcaporaso and I chatted about this metric rename. We'd prefer to leave it as `observed_otus` because the name has already changed (in QIIME 1/PyCogent, it used to be `observed_species`, then changed to `observed_otus` somewhere around QIIME 1.8). Using the current name also lets us link to the skbio diversity metric docs since the names correspond 1-1.

We discussed making an alias but then users might be confused about how the two metrics compare. Right now we don't have a great way to document each metric so it seems safest/easiest on users if we keep the current (less-than-ideal) name. If we do change it in the future, we'd probably call it `observed_features` to keep the naming scheme intact.

q2_diversity/_alpha/_method.py
```diff
@@ -36,12 +36,20 @@ def alpha_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
-                otu_ids=feature_ids,
-                tree=phylogeny)
+    try:
+        result = skbio.diversity.alpha_diversity(metric=metric,
+                                                 counts=counts,
+                                                 ids=sample_ids,
+                                                 otu_ids=feature_ids,
+                                                 tree=phylogeny)
+    except skbio.tree.MissingNodeError as e:
+        if 'otu_ids' in str(e):
```
jairideout: Even safer, what if we did a `str.replace("``otu_ids``", "feature IDs")` and `str.replace("``tree``", "``phylogeny``")`, and then raised a new `MissingNodeError` with the updated error message? That way we aren't assuming much about the type of `MissingNodeError` (it's possible that "otu_ids" appears in a different type of `MissingNodeError`), and we've rewritten parts of the error message to match the QIIME 2 API instead of scikit-bio (e.g. "phylogeny" instead of "tree").

Same comment re: beta diversity handling below.

q2_diversity/tests/test_alpha.py
```diff
@@ -27,9 +27,10 @@ def test_alpha(self):
         t = biom.Table(np.array([[0, 1, 3], [1, 1, 2]]),
                        ['O1', 'O2'],
                        ['S1', 'S2', 'S3'])
-        actual = alpha(table=t, metric='observed_otus')
+        actual = alpha(table=t, metric='feature_counts')
```
jairideout: Can you add a couple quick tests for alpha and beta diversity that asserts the error message is rewritten?

q2_diversity/_alpha/_method.py
jakereps: Reverted!

q2_diversity/tests/test_alpha.py
jakereps: Done!

jakereps:
Thanks @jairideout! Everything should be addressed!

jairideout:
👍 , thanks @jakereps!

PR #93, opened 2017-02-27
TST: adds coverage testing and reporting to Travis CI

gregcaporaso:


jairideout:
Adding *do not merge* label until the `pip install -e .` issue is sorted out.

gregcaporaso:
This latest commit seems to fix the issue (locally at least - travis doesn't seem to have started yet). The issue was that the `tests` directory was missing a ``__init__.py`` file, and this seems to be required for `py.test --cov` to function correctly.

gregcaporaso:
Travis/Coverage are passing now, and the coverage report is correct on coveralls.

jairideout:
Thanks @gregcaporaso! The missing `__init__.py` would have affected `nosetests` as well. It's possible these tests weren't being executed before, but I haven't confirmed that. Thanks for fixing!

PR #96, opened 2017-04-05
ENH: adds optional pairwise beta group significance tests

### LINKED ISSUE #50: add pairwise tests for beta-group-significance
this came up in the iceland workshop

gregcaporaso:
fixes #50 

[Here's](https://dl.dropboxusercontent.com/u/2868868/temp/unweighted-unifrac-body-site-pairwise-significance.qzv) an example .qzv with the pairwise tests.

I've confirmed that the .qzv looks right when computed with ANOSIM (some of the descriptions of values change), and when it is run without the pairwise tests. 

Pairwise tests are off by default since they can be very slow.

### DIFF
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -191,16 +191,18 @@ def beta_group_significance(output_dir: str,
             pairwise_results.append([group1_id,
                                      group2_id,
                                      pairwise_result['sample size'],
+                                     permutations,
                                      pairwise_result['test statistic'],
                                      pairwise_result['p-value']])
-        columns = ['Group 1', 'Group 2', 'Sample size',
+        columns = ['Group 1', 'Group 2', 'Sample size', 'Permutations',
                    result['test statistic name'], 'p-value']
         pairwise_results = pd.DataFrame(pairwise_results, columns=columns)
         pairwise_results.set_index(['Group 1', 'Group 2'], inplace=True)
         pairwise_results['q-value'] = multipletests(
             pairwise_results['p-value'], method='fdr_bh')[1]
         pairwise_results.sort_index(inplace=True)
-        pairwise_path = os.path.join(output_dir, 'pairwise.csv')
+        pairwise_path = os.path.join(
+            output_dir, '%s-pairwise.csv' % method)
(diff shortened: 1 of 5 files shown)


jairideout:
travis failures

jairideout:
@gregcaporaso, I looked into the travis failures and it's the q2lint issue @jakereps brought up on Slack (too old version of Node on travis). For now let's ignore that error and I'll confirm locally that q2lint passes.

jairideout:
Confirmed that q2lint passes locally. @gregcaporaso, can you try adding `dist: trusty` below the `sudo: false` line in .travis.yml?

q2_diversity/_beta/_visualizer.py
```diff
@@ -159,15 +172,51 @@ def beta_group_significance(output_dir: str,
...
+                                     pairwise_result['test statistic'],
+                                     pairwise_result['p-value']])
+        columns = ['Group 1', 'Group 2', 'Sample size',
+                   result['test statistic name'], 'p-value']
+        pairwise_results = pd.DataFrame(pairwise_results, columns=columns)
+        pairwise_results.set_index(['Group 1', 'Group 2'], inplace=True)
+        pairwise_results['q-value'] = multipletests(
+            pairwise_results['p-value'], method='fdr_bh')[1]
+        pairwise_results.sort_index(inplace=True)
+        pairwise_path = os.path.join(output_dir, 'pairwise.csv')
```
jairideout: It'd also be useful to include a "number of permutations" column in the pairwise results so that the p-values can be interpreted/reported without needing the full visualization and its global test results.

jairideout:
I was poking through the example results you posted (BodySite category), and the p-values vs. q-values look a little odd to me:

screen shot 2017-04-06 at 10 03 32 am

The q-values appear to be the same as the p-values, with some extra precision added. @gregcaporaso can you confirm that these are the expected results? I'm not familiar with statsmodels' `fdr_bh` correction but I would expect the q-values to be noticeably higher since 6 tests were performed.

gregcaporaso:
@jairideout, addressed your comments. I computed these q-values with statsmodels directly (independent of this code) and an online calculator, and I got the same values so these should be correct.

PR #109, opened 2017-06-13
ENH: Support for negating distance matrix filtering

maxvonhippel:
This feature replicates the `exclude_ids` feature [recently implemented in q2-feature-table](https://github.com/qiime2/q2-feature-table/pull/112).  When `exclude_ids=True`, the samples with ids matching those satisfying the `metadata` (or optionally the `where` clause on the `metadata`) are excluded from rather than included in the DistanceMatrix that is returned by `_filter`.  `False` by default.

### DIFF
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -12,13 +12,22 @@ import qiime2
 
 def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
                            metadata: qiime2.Metadata,
-                           where: str=None) -> skbio.DistanceMatrix:
+                           where: str=None,
+                           exclude_ids: bool=False) -> skbio.DistanceMatrix:
     ids_to_keep = metadata.ids(where=where)
     # NOTE: there is no guaranteed ordering to output distance matrix because
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
     try:
-        return distance_matrix.filter(ids_to_keep, strict=False)
+        filtered = distance_matrix.filter(ids_to_keep, strict=False)
+        if exclude_ids is True:
+            ids_to_keep = set(distance_matrix.ids) - set(filtered.ids)
+            filtered = distance_matrix.filter(ids_to_keep, strict=False)
+        return filtered
     except skbio.stats.distance.DissimilarityMatrixError:
-        raise ValueError(
-            "All samples were filtered out of the distance matrix.")
+        if (exclude_ids and ids_to_keep and
+                not bool(set(ids_to_keep) & set(distance_matrix.ids))):
+            return distance_matrix
+        else:
+            raise ValueError("All samples were filtered out of the "
+                             "distance matrix.")
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -20,7 +20,7 @@ def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
     # order.
     try:
         filtered = distance_matrix.filter(ids_to_keep, strict=False)
-        if exclude_ids is True:
+        if exclude_ids:
             ids_to_keep = set(distance_matrix.ids) - set(filtered.ids)
             filtered = distance_matrix.filter(ids_to_keep, strict=False)
         return filtered
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -19,15 +19,9 @@ def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
     try:
-        filtered = distance_matrix.filter(ids_to_keep, strict=False)
         if exclude_ids:
-            ids_to_keep = set(distance_matrix.ids) - set(filtered.ids)
-            filtered = distance_matrix.filter(ids_to_keep, strict=False)
-        return filtered
+            ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
+        return distance_matrix.filter(ids_to_keep, strict=False)
     except skbio.stats.distance.DissimilarityMatrixError:
-        if (exclude_ids and ids_to_keep and
-                set(ids_to_keep).isdisjoint(distance_matrix.ids)):
-            return distance_matrix
-        else:
-            raise ValueError("All samples were filtered out of the "
-                             "distance matrix.")
+        raise ValueError("All samples were filtered out of the "
+                         "distance matrix.")
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -15,11 +15,11 @@ def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
                            where: str=None,
                            exclude_ids: bool=False) -> skbio.DistanceMatrix:
     ids_to_keep = metadata.ids(where=where)
+    if exclude_ids:
+        ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
     # NOTE: there is no guaranteed ordering to output distance matrix because
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
-    if exclude_ids:
-        ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
     try:
         return distance_matrix.filter(ids_to_keep, strict=False)
     except skbio.stats.distance.DissimilarityMatrixError:
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -26,7 +26,7 @@ def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
         return filtered
     except skbio.stats.distance.DissimilarityMatrixError:
         if (exclude_ids and ids_to_keep and
-                not bool(set(ids_to_keep) & set(distance_matrix.ids))):
+                set(ids_to_keep).isdisjoint(distance_matrix.ids)):
             return distance_matrix
         else:
             raise ValueError("All samples were filtered out of the "
--- a/q2_diversity/_filter.py
+++ b/q2_diversity/_filter.py
@@ -18,10 +18,10 @@ def filter_distance_matrix(distance_matrix: skbio.DistanceMatrix,
     # NOTE: there is no guaranteed ordering to output distance matrix because
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
+    if exclude_ids:
+        ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
     try:
-        if exclude_ids:
-            ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
         return distance_matrix.filter(ids_to_keep, strict=False)
     except skbio.stats.distance.DissimilarityMatrixError:
-        raise ValueError("All samples were filtered out of the "
-                         "distance matrix.")
+        raise ValueError(
+            "All samples were filtered out of the distance matrix.")
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -233,7 +233,7 @@ plugin.methods.register_function(
                  'also in the input distance matrix will be retained.',
         'exclude_ids': 'If `True`, returns the filtered distance matrix '
                        'containing the set of samples which occur in the '
-                       'feature table but not in the metadata (or the '
+                       'distance matrix but not in the metadata (or the '
                        'filtered metadata, if invoked in conjunction with '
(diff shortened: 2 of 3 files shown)


maxvonhippel:
One thing worth discussing: as is, the code behaves differently for zero filtering via `metadata` and zero filtering via `where`.  If you filter nothing with `metadata` and set `exclude_ids=True`, you get your original DistanceMatrix back, which seems sensible to me.

````PYTHON
# metadata filter nothing -> exclude nothing
df = pd.DataFrame({'Subject': ['subject-1'],
                   'SampleType': ['tongue']},
                   index=['S2000'])
metadata = qiime2.Metadata(df)

filtered = filter_distance_matrix(dm, metadata,
                                  where=None,
                                  exclude_ids=True)
self.assertEqual(self._sorted(filtered), dm)
````

However, if you write a `where` clause that can't filter anything, you get a `ValueError`:

````PYTHON
# where filter nothing -> exclude nothing
df = pd.DataFrame({'Subject': ['subject-1', 'subject-1', 'subject-2'],
                   'SampleType': ['gut', 'tongue', 'gut']},
                  index=['S1', 'S2', 'S3'])
metadata = qiime2.Metadata(df)

with self.assertRaisesRegex(ValueError,
                            "Selection.*IDs.*failed.*query.*"):
    filter_distance_matrix(dm, metadata,
                           where="SampleType='toe",
                           exclude_ids=True)
````

The actual error looks like this:

````BASH
E           AssertionError: "Selection*IDs*failed*query*" does not match "Selection of IDs failed with query:
E            SELECT "7b7da3b4-4f56-4fa8-9f27-41330de78086" FROM metadata WHERE SampleType='toe GROUP BY "7b7da3b4-4f56-4fa8-9f27-41330de78086" ORDER BY "7b7da3b4-4f56-4fa8-9f27-41330de78086";"

q2_diversity/tests/test_filter.py:196: AssertionError
````

I propose that, if the above `where` ValueError is raised, it makes sense to just not filter anything, because clearly that `where` clause isn't filtering anything.  This is how I think most ORMs tend to work.  For example, in SQL, if I did something like

````SQL
SELECT * FROM groovy_table WHERE awesome_id="baloney"
````

.... but only had ids "hotdog", "keyboard", and "dog" in `groovy_table`, I'd just get nothing returned.  I'm fairly certain it would not error in this case, as our code currently does.

Thoughts?

**Comment by thermokarst** [comment]
Just a quick note @maxvonhippel, there are some issues with your SQL in the comment above that appear to be confounding whatever it is you are [reporting on](https://github.com/qiime2/q2-diversity/pull/109#issuecomment-308272126).

e.g.:
```python
filter_distance_matrix(dm, metadata, where="SampleType='toe", exclude_ids=True)
```

is generating the following incomplete SQL:

```sql
SELECT "7b7da3b4-4f56-4fa8-9f27-41330de78086" FROM metadata WHERE SampleType='toe GROUP BY "7b7da3b4-4f56-4fa8-9f27-41330de78086" ORDER BY "7b7da3b4-4f56-4fa8-9f27-41330de78086";
```

Note the missing closing single quote around the word `toe`.

Revised:
```python
filter_distance_matrix(dm, metadata, where="SampleType='toe'", exclude_ids=True)
```

Running that:
```
---------------------------------------------------------------------------
DissimilarityMatrixError                  Traceback (most recent call last)
~/caporaso/qiime2/q2-diversity/q2_diversity/_filter.py in filter_distance_matrix(distance_matrix, metadata, where, exclude_ids)
     21     try:
---> 22         filtered = distance_matrix.filter(ids_to_keep, strict=False)
     23         if exclude_ids is True:

~/miniconda3/envs/q2dev-2017.6/lib/python3.5/site-packages/skbio/stats/distance/_base.py in filter(self, ids, strict)
    367         filtered_data = self._data[idxs][:, idxs]
--> 368         return self.__class__(filtered_data, ids)
    369

~/miniconda3/envs/q2dev-2017.6/lib/python3.5/site-packages/skbio/stats/distance/_base.py in __init__(self, data, ids)
    106
--> 107         self._validate(data, ids)
    108

~/miniconda3/envs/q2dev-2017.6/lib/python3.5/site-packages/skbio/stats/distance/_base.py in _validate(self, data, ids)
    867         """
--> 868         super(DistanceMatrix, self)._validate(data, ids)
    869

~/miniconda3/envs/q2dev-2017.6/lib/python3.5/site-packages/skbio/stats/distance/_base.py in _validate(self, data, ids)
    665         if 0 in data.shape:
--> 666             raise DissimilarityMatrixError("Data must be at least 1x1 in "
    667                                            "size.")

DissimilarityMatrixError: Data must be at least 1x1 in size.

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-13-9980a2e5e616> in <module>()
----> 1 filter_distance_matrix(dm, metadata, where="\"SampleType\"='toe'", exclude_ids=True)

~/caporaso/qiime2/q2-diversity/q2_diversity/_filter.py in filter_distance_matrix(distance_matrix, metadata, where, exclude_ids)
     30             return distance_matrix
     31         else:
---> 32             raise ValueError("All samples were filtered out of the "
     33                              "distance matrix.")

ValueError: All samples were filtered out of the distance matrix.
```

This `ValueError` makes a lot more sense than the error you reported above due to malformed SQL --- it clearly reports that all samples were filtered.
Thanks!

maxvonhippel:
@thermokarst good catch - that makes a lot more sense.

So in this case, if I understand correctly, it sounds like the behavior of the code as of this PR is what we want/expect?

thermokarst:
@maxvonhippel: I'm not sure, I will leave that call up to you. If you think it is ready to be reviewed just let me know and I can take a look!

maxvonhippel:
@thermokarst this is ready for review.

I am still ambivalent about the message for [the `where` clause](https://github.com/qiime2/q2-diversity/pull/109/files#diff-2da5bf068d9eb56eb1591bb87cfef6b9R230).  I think it might make sense to modify it to make clear that exclusion is possible.  For example:

> SQLite WHERE clause specifying sample metadata criteria that must be met to be included (or excluded, if `exclude_ids` is True) in the filtered distance matrix. If not provided, all samples in `metadata` that are also in the input distance matrix will be retained (or dropped, if `exclude_ids` is True).

My wording above is admittedly awkward, but I feel like it is also slightly more accurate.  That said, we did not say anything about exclusion in the similar PR on `q2-feature-table`.

Anyway, thank you for reviewing this, and please let me know if you think the `where` clause description should be changed or if it makes sense as-is.

q2_diversity/tests/test_filter.py
```diff
@@ -135,6 +135,105 @@ def test_with_where_some_filtered(self):
...
+        # where filter nothing -> exclude nothing
+        df = pd.DataFrame({'Subject': ['subject-1', 'subject-1', 'subject-2'],
+                           'SampleType': ['gut', 'tongue', 'gut']},
+                          index=['S1', 'S2', 'S3'])
+        metadata = qiime2.Metadata(df)
+
+        with self.assertRaisesRegex(ValueError,
+                                    "Selection.*IDs.*failed.*query.*"):
+            filter_distance_matrix(dm, metadata,
+                                   where="SampleType='toe",
```
thermokarst: Missing a closing quote around `toe`.

q2_diversity/tests/test_filter.py
```diff
@@ -135,6 +135,105 @@ def test_with_where_some_filtered(self):
...
+            filter_distance_matrix(dm, metadata, where=None, exclude_ids=True)
+
+        # where filter nothing -> exclude nothing
+        df = pd.DataFrame({'Subject': ['subject-1', 'subject-1', 'subject-2'],
+                           'SampleType': ['gut', 'tongue', 'gut']},
+                          index=['S1', 'S2', 'S3'])
+        metadata = qiime2.Metadata(df)
+
+        with self.assertRaisesRegex(ValueError,
+                                    "Selection.*IDs.*failed.*query.*"):
```
thermokarst: This needs to be updated once the `toe`-quoting issue is fixed (see below).

q2_diversity/_filter.py
```diff
@@ -12,13 +12,22 @@
...
-                           where: str=None) -> skbio.DistanceMatrix:
+                           where: str=None,
+                           exclude_ids: bool=False) -> skbio.DistanceMatrix:
     ids_to_keep = metadata.ids(where=where)
     # NOTE: there is no guaranteed ordering to output distance matrix because
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
     try:
-        return distance_matrix.filter(ids_to_keep, strict=False)
+        filtered = distance_matrix.filter(ids_to_keep, strict=False)
```
thermokarst: We can refactor this to avoid computing `filtered` twice, right? We can use the same strategy as in [q2-feature-table](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_filter.py#L48-L49).

```python
if exclude_ids:
    ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
return distance_matrix.filter(ids_to_keep, strict=False)
```

q2_diversity/_filter.py
```diff
@@ -12,13 +12,22 @@
...
+        if exclude_ids:
+            ids_to_keep = set(distance_matrix.ids) - set(filtered.ids)
+            filtered = distance_matrix.filter(ids_to_keep, strict=False)
+        return filtered
     except skbio.stats.distance.DissimilarityMatrixError:
-        raise ValueError(
-            "All samples were filtered out of the distance matrix.")
+        if (exclude_ids and ids_to_keep and
```
thermokarst: Can you explain what this fall-through is for? I think I have an idea, but maybe we need a brief comment in the code to explain what this is for. As well, I don't think I see any tests that cover this check --- can you add one if that is the case, or point out to me how you are testing this?

q2_diversity/plugin_setup.py
```diff
@@ -227,7 +230,12 @@
         'where': 'SQLite WHERE clause specifying sample metadata criteria '
                  'that must be met to be included in the filtered distance '
                  'matrix. If not provided, all samples in `metadata` that are '
-                 'also in the input distance matrix will be retained.'
+                 'also in the input distance matrix will be retained.',
+        'exclude_ids': 'If `True`, returns the filtered distance matrix '
+                       'containing the set of samples which occur in the '
+                       'feature table but not in the metadata (or the '
```
thermokarst: `s/feature table/distance matrix`

q2_diversity/_filter.py
maxvonhippel: You're right, thank you.  And this also solves the problem which my elaborate and unnecessary code in the `except` was attempting to handle.

q2_diversity/_filter.py
maxvonhippel: This is unnecessary, it was because I failed to see that I could just use `set(ids_to_keep)` in the first place.  As a result, in the albeit faulty `toe` example, I was concerned with the empty matrix exception happening before I got to the exclusion step.  But now that I am just subtracting the ids in the first place without creating `filtered` twice this step is not needed.

maxvonhippel:
Ok, this is ready for another pass of review.  Thanks!

q2_diversity/_filter.py
```diff
@@ -12,13 +12,16 @@
...
                            metadata: qiime2.Metadata,
-                           where: str=None) -> skbio.DistanceMatrix:
+                           where: str=None,
+                           exclude_ids: bool=False) -> skbio.DistanceMatrix:
     ids_to_keep = metadata.ids(where=where)
     # NOTE: there is no guaranteed ordering to output distance matrix because
     # `ids_to_keep` is a set, and `DistanceMatrix.filter` uses its iteration
     # order.
     try:
+        if exclude_ids:
```
thermokarst: This is a bit on the stylistic side, but I think it makes sense to move this check up to line 18 above. I suggest this because this code (`ids_to_keep = ...`) doesn't make much sense to include in the `try/except` block, and in particular, you aren't checking for any potential errors that **could** be raised in L23 (I'm not suggesting you do). Make sense?

q2_diversity/_filter.py
```diff
@@ -12,13 +12,16 @@
...
+            ids_to_keep = set(distance_matrix.ids) - set(ids_to_keep)
         return distance_matrix.filter(ids_to_keep, strict=False)
     except skbio.stats.distance.DissimilarityMatrixError:
-        raise ValueError(
-            "All samples were filtered out of the distance matrix.")
+        raise ValueError("All samples were filtered out of the "
+                         "distance matrix.")
```
thermokarst: Also stylistic, but is this change necessary? Could it confuse someone looking at the git history in the future?

maxvonhippel:
Ok, thanks @thermokarst !  This is ready for another pass

maxvonhippel:
Thanks @jairideout , just implemented both of your suggestions!

PR #116, opened 2017-07-20
ENH: Allow beta methods to utilize parallel jobs.

### LINKED ISSUE #108: Expose, or support, sklearn.metrics.pairwise_distances in beta/alpha diversity methods
Currently the `beta` and `alpha` diversity methods use [`scipy.spatial.distance.pdist`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html) (through the default method for `skbio`). `sklearn` offers a pairwise distance function ([`sklearn.metrics.pairwise_distances`](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html#sklearn.metrics.pairwise_distances)) that utilizes the same scipy method internally, but also allows parallelization (and adds a few other metrics of their own). 

Here's a small example benchmark:
image

Could this alternative function be either swapped as the default function used, or made available as an option of some sort, where the existing behavior could even remain the same just a `n_jobs` parameter added that defaults to 1? 

It could possibly be made as an optional parameter where the description disclaims that it 'requires sklearn' or something in the parameter description, without requiring `sklearn` to be added as another dependency, and the import can be wrapped in a try/except ImportError block.

jakereps:
Resolves #108 

Curious on the parameter descriptions. Since it is straight from sklearn I just ported it over, and mentioned on `core_metrics` that it only applies to the `beta`s.

### DIFF
--- a/q2_diversity/_beta/_method.py
+++ b/q2_diversity/_beta/_method.py
@@ -29,11 +29,13 @@ def non_phylogenetic_metrics():
 
 
 def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
-                      metric: str)-> skbio.DistanceMatrix:
+                      metric: str, n_jobs: int=1)-> skbio.DistanceMatrix:
     if metric not in phylogenetic_metrics():
         raise ValueError("Unknown phylogenetic metric: %s" % metric)
     if table.is_empty():
         raise ValueError("The provided table object is empty")
+    if n_jobs != 1 and metric == 'weighted_unifrac':
+        raise ValueError("Weighted UniFrac is not parallelizable")
 
     counts = table.matrix_data.toarray().astype(int).T
     sample_ids = table.ids(axis='sample')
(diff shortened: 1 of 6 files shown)


q2_diversity/_beta/_method.py
```diff
@@ -11,6 +11,7 @@
 import skbio
 import skbio.diversity
 import skbio.tree
+import sklearn.metrics
```
jakereps: I was looking everywhere for the requirements, haha! Haven't added a dependency since the CD refactors, so was really confused on why they weren't in the `setup.py` anymore.

jairideout: Sorry for the confusion -- we have a private (for now) draft doc that's on the forum detailing some of these qiime2 package requirements. I'll share that with you on Slack.

jakereps:
Thanks @jairideout! Sorry on the delay. I've addressed your comments.

jairideout:
It looks like `n_jobs` got dropped from `beta_phylogenetic`, can you add that back? `beta_phylogenetic` can use `n_jobs` for other metrics except for weighted UniFrac.

Comment from previous review:

> For now, can you have core-metrics avoid using n-jobs with weighted UniFrac, and also have beta-phylogenetic raise an error if weighted UniFrac and non-default n-jobs is used? Can you also update the parameter descriptions on core-metrics and beta-phylogenetic to describe this special case behavior?

jakereps:
Oops, I think I misread it the first round through that phylo wouldn't work at all. Added it back! Thanks @jairideout!

PR #137, opened 2017-09-06
Match metadata to alpha diversity data before type inference

ElDeveloper:


jairideout:
Closing this PR now that Metadata has been overhauled and column typing works differently. The Metadata changes haven't made it into a release yet, that's scheduled for 2018.2, but latest master branches and qiime2 dev environments have access to the new Metadata support.

The way it works now is that metadata column types are either 1) inferred when the file is loaded; or 2) are explicitly typed by modifying the metadata file to use a "comment directive" (see the second row in the [Moving Pictures sample metadata](https://docs.google.com/spreadsheets/d/1LpGUOdXwCs6m4xR9DoB8yOl_22v1hczjZHyA41DhDXg/edit#gid=0) for an example), or by using the `column_types` parameter with `qiime2.Metadata.load()`. Once the file is loaded, a column's type is fixed (i.e. a `numeric` column stays `numeric` regardless of filtering; `categorical` stays `categorical` as well). We decided to have fixed column types because otherwise a user wouldn't have control over what their column types are after filtering happens within a plugin's method/visualizer. For example, if a user has a categorical column consisting of strings and numbers, and the plugin method/visualizer filters that down to only numbers, the user wouldn't have a way to say "no, please keep that column categorical and don't recast".

Note: we plan to have a `qiime2.Metadata.recast()` method (method name TBD) that will allow Artifact API users to perform filtering and have column types recast, but we would encourage plugin developers to avoid using `recast()` in a method/visualizer (the `recast()` method will raise a warning to help catch improper usage within a plugin). I don't have an ETA on when `recast()` will be available though.

Let us know how the Metadata column typing works out for you!

PR #143, opened 2017-09-20
Adds a method to use a high performance implementation of UniFrac

ElDeveloper:
This PR adds the necessary code to use the implementation of UniFrac in https://github.com/biocore/unifrac . This method is recommended for large datasets.

As a comparison, for a table with ~120 samples and a tree with 20K tips running time is: 2 minutes and 27 seconds with the current implementation, and 2 seconds with the new implementation. @wasade has been benchmarking this more thoroughly but, thought this served as a nice and small comparison.

As discussed at the developer's meeting, this method does not replace beta_phylogenetic, as it is still fairly new, instead it is provided as an alternative (beta_phylogenetic_hpc) that users can test out.

Currently, there's a few things missing in this PR:

- [x] Tests, it would make more sense to add a few files to the repo, rather than having to create those on the fly, but I am not sure if there's a preferred way to do this.
- [x] Deployment, this depends on https://github.com/biocore/unifrac which does not yet have a conda recipe (@wasade is working on this).
- [x] Citation information, it would be nice to include more information about the specific papers that should be cited for each metric. We previously had this when this was its own plugin, [see here](https://github.com/biocore/unifrac/blob/a9e5118b721e45e36afb788173d54c0985801636/q2_state_unifrac/plugin_setup.py).
- [ ] Input validation, there's a few validations that are currently not being performed as part of https://github.com/biocore/unifrac but that are made here or in skbio, for example see (https://github.com/biocore/unifrac/issues/37, https://github.com/biocore/unifrac/issues/36, https://github.com/biocore/unifrac/issues/35, https://github.com/biocore/unifrac/issues/34 and https://github.com/biocore/unifrac/issues/33 )


wasade:
Thanks, @ElDeveloper! One additional point, we include support for Generalized UniFrac and Variance Adjusted UniFrac. We also support meta UniFrac but it assumes variadic arguments support (qiime2/qiime2#234); it may be nice to put in a near term work around though.

jairideout:
> Tests, it would make more sense to add a few files to the repo, rather than having to create those on the fly, but I am not sure if there's a preferred way to do this.

Take a look at some of the other QIIME 2 repos for examples of adding test data files -- we have a custom `unittest.TestCase` subclass that provides a helper for locating test data. I think q2-types is a good example -- let us know if you run into issues. Please try to keep the files small.

> Citation information, it would be nice to include more information about the specific papers that should be cited for each metric. We previously had this when this was its own plugin, see here.

Citation support is currently limited to free text, so you can add whatever citations you want to the plugin using the `citation_text` field (I don't recommend putting citation info in the `description` field like you have in your linked example).

ElDeveloper:
Thanks for the guidance Jai, I'll put these in place.

ElDeveloper:
@jairideout, I've added the citation text and tests. 👍 This should be contingent on @wasade publishing a conda package, but is otherwise good to review.

wasade:
@ElDeveloper, would it be possible to expose `bypass_tips`, a.k.a. fast mode? As a parameter description: "Compute will be reduced by ~50% with this option yielding an approximate result. The option disregards computing UniFrac at the tips of the phylogeny and instead only computes it on the internal nodes. In a bifurcating tree, the tips correspond to approximately 50% of the vertices in the tree, which is how the compute time reduction is achieved. The use of this parameter is analogous in concept to switching from 99% OTUs to 97% OTUs."

ElDeveloper:
Tests are passing 🎉

ebolyen:
Looks like [it worked](https://github.com/qiime2/environment-files/commit/2e173e3ced976ef934f6e96a2f0cf99542a13b5a)! We've got new environment files to test with 🎉!

wasade:
That's excellent!!

On Sep 23, 2017 11:49, "Evan Bolyen" <notifications@github.com> wrote:

Looks like it worked
<https://github.com/qiime2/environment-files/commit/2e173e3ced976ef934f6e96a2f0cf99542a13b5a>!
We've got new environment files to test with 🎉!

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/qiime2/q2-diversity/pull/143#issuecomment-331662284>,
or mute the thread

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,65 @@
...
+                 " for all pairs of samples in a feature table. This "
+                 "implementation is recommended for large datasets, otherwise "
+                 "the results are identical to beta_phylogenetic.\n\n"
+                 "This method is an implementation of the Strided State "
+                 "UniFrac algorithm. Multiple variants of the UniFrac metric "
+                 "are available, including Generalized UniFrac (Chen et al. "
+                 "2012), Variance Adjusted UniFrac (Chang et al. 2011), "
+                 "as well as Weighted normalized and unnormalized UniFrac "
+                 "(Lozupone et al. 2007), unweighted UniFrac "
+                 "(Lozupone et al. 2005), and metagenomic UniFrac "
```
ebolyen: I don't think metagenomic UniFrac is exposed at the moment.

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +69,27 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
+                          alpha=1.0,
+                          bypass_tips: bool=False) -> skbio.DistanceMatrix:
+    if metric == 'unweighted_unifrac':
+        f = unifrac.unweighted
+    elif metric == 'weighted_unnormalized_unifrac':
+        f = unifrac.weighted_unnormalized
+    elif metric == 'weighted_normalized_unifrac':
+        f = unifrac.weighted_normalized
+    elif metric == 'generalized_unifrac':
+        f = partial(unifrac.generalized, alpha=alpha)
```
ebolyen: This is fine, but a dictionary lookup might be cleaner (that's usually what we do).

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +69,27 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
     return results
 
 
+def beta_phylogenetic_hpc(table: BIOMV210Format, phylogeny: NewickFormat,
+                          metric: str, n_jobs: int=1,
+                          variance_adjusted: bool=False,
+                          alpha=1.0,
```
ebolyen: I don't have a strong recommendation, but it might be nice to error if `alpha` is passed when the metric isn't generalized UniFrac. That would require you set the default to something like `None` though, which definitely obfuscates things a bit.

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +69,27 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
     return results
 
 
+def beta_phylogenetic_hpc(table: BIOMV210Format, phylogeny: NewickFormat,
```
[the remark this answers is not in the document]
ElDeveloper: Great suggestion, changed to alt!

q2_diversity/_beta/_method.py
ElDeveloper: I followed your suggestion here.

q2_diversity/_beta/_method.py
ElDeveloper: It made things cleaner, though we needed a bit of special casing though 👍

q2_diversity/plugin_setup.py
```diff
@@ -38,7 +38,19 @@
...
+                   '10.1128/AEM.71.12.8228-8235.2005.'
+                   'Weighted UniFrac: '
+                   'Lozupone et al. 2007 Appl Environ Microbiol; DOI: '
+                   '10.1128/AEM.01996-06.'
+                   'Variance adjusted UniFrac: '
+                   'Chang et al. BMC Bioinformatics 2011 '
+                   'https://doi.org/10.1186/1471-2105-12-118.'
+                   'Generalized UniFrac: '
+                   'Chen et al. 2012 Bioinformatics; DOI: '
+                   '10.1093/bioinformatics/bts342')
```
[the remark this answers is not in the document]
ElDeveloper: I couldn't see any changes with the new lines I added.  ¯\_(ツ)_/¯

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,65 @@
...
+
+plugin.methods.register_function(
+    function=q2_diversity.beta_phylogenetic_hpc,
+    inputs={'table': FeatureTable[Frequency],
+            'phylogeny': Phylogeny[Rooted]},
+    parameters={'metric': Str % Choices(['unweighted_unifrac',
+                                         'weighted_unnormalized_unifrac',
+                                         'weighted_normalized_unifrac',
+                                         'generalized_unifrac']),
+                'n_jobs': Int,
```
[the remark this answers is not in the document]
ElDeveloper: Negative numbers are not supported, so this is great!

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,65 @@
...
+                'variance_adjusted': Bool,
+                'alpha': Float, 'bypass_tips': Bool},
```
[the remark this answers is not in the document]
ElDeveloper: Not sure what this means, but done.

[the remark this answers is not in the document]
ElDeveloper: Great, I had previously tried `Float % Range(0, 1)` but that wasn't working, I'm glad this works.

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,65 @@
...
+                      'correspond to the feature identifiers in the table. '
+                      'This tree can contain tip ids that are not present in '
+                      'the table, but all feature ids in the table must be '
+                      'present in this tree.')
+    },
+    parameter_descriptions={
+        'metric': 'The beta diversity metric to be computed.',
+        'n_jobs': 'The number of workers to use.',
+        'variance_adjusted': ('Perform variance adjustment based on Chang et '
+                              'al. BMC Bioinformatics 2011'),
```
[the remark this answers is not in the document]
ElDeveloper: I've added an explanation, let me know if it makes sense, also CCing @wasade to keep him in the loop.

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,65 @@
...
+                  'proportions are dichotomized.'),
+        'bypass_tips': ('Compute will be reduced by ~50% with this option '
+                        'yielding an approximate result. The option '
+                        'disregards computing UniFrac at the tips of the '
+                        'phylogeny and instead only computes it on the '
+                        'internal nodes. In a bifurcating tree, the tips '
+                        'correspond to approximately 50% of the vertices in '
+                        'the tree, which is how the compute time reduction is'
+                        ' achieved. The use of this parameter is analogous '
+                        'in concept to switching from 99% OTUs to 97% OTUs.')
```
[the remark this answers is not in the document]
ElDeveloper: That's much more concise, thanks.

q2_diversity/plugin_setup.py
ElDeveloper: Good catch.

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +76,31 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
     return results
 
 
+def beta_phylogenetic_alt(table: BIOMV210Format, phylogeny: NewickFormat,
```
wasade: Would it be possible to politely catch and err if the `table` has < 4 samples? That's an edge case we haven't handled yet behind the scenes

q2_diversity/plugin_setup.py
```diff
@@ -38,15 +38,27 @@
...
+                   '10.1093/bioinformatics/bts342')
 )
 
 plugin.methods.register_function(
     function=q2_diversity.beta_phylogenetic,
     inputs={'table': FeatureTable[Frequency],
             'phylogeny': Phylogeny[Rooted]},
     parameters={'metric': Str % Choices(beta.phylogenetic_metrics()),
-                'n_jobs': Int},
+                'n_jobs': Int % Range(1, None)},
```
wasade: This is minor, and ultimately whatever fits better into the surrounding terminology is preferred, but it is more correct to describe this as threads.

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,58 @@
...
+                      'This tree can contain tip ids that are not present in '
+                      'the table, but all feature ids in the table must be '
+                      'present in this tree.')
+    },
+    parameter_descriptions={
+        'metric': 'The beta diversity metric to be computed.',
+        'n_jobs': 'The number of workers to use.',
+        'variance_adjusted': ('Perform variance adjustment based on Chang et '
+                              'al. BMC Bioinformatics 2011. Weights distances '
+                              'by the number of tips evaluated for each '
```
wasade: It weights based on the proportion of the relative abundance represented between the samples at a given node under evaluation

q2_diversity/plugin_setup.py
```diff
@@ -68,6 +80,58 @@
...
+                              'comparison.'),
+        'alpha': ('This parameter is only used when the choice of metric is '
+                  'generalized_unifrac. The value of alpha controls importance'
+                  ' of sample proportions. 1.0 is weighted normalized UniFrac.'
+                  ' 0.0 is close to unweighted UniFrac, but only if the sample'
+                  ' proportions are dichotomized.'),
+        'bypass_tips': ('In a bifurcating tree, the tips make up about 50% of '
+                        'the nodes in a tree. By ignoring them, specificity '
+                        'can be traded for reduced compute time.')
```
wasade: Perhaps add: "time. This has the effect of collapsing the phylogeny, and is analogous in concept to moving from 99% to 97% OTUs"?

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +76,31 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
+                          bypass_tips: bool=False) -> skbio.DistanceMatrix:
+
+    metrics = phylogenetic_metrics_alt_dict()
+
+    if metric not in metrics:
+        raise ValueError("Unknown metric: %s" % metric)
+
+    if alpha is not None and metric != 'generalized_unifrac':
+        raise ValueError('The alpha parameter is only allowed when the choice'
+                         ' of metric is generalized_unifrac')
```
ebolyen: The "opposite" of this causes `unifrac` to raise an error, but the user is too far removed to be able to pinpoint the cause.

So we should add a branch to either set `alpha=1` when `alpha == None` or to explain that `alpha` needs to be set. Either should work.

q2_diversity/_beta/_method.py
```diff
@@ -63,6 +76,31 @@ def beta_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
...
+
+    if metric == 'generalized_unifrac':
+        f = partial(metrics[metric], alpha=alpha)
+    else:
+        f = metrics[metric]
+
+    # unifrac processes tables and trees should be filenames
+    return f(str(table), str(phylogeny), threads=n_jobs,
```
ebolyen: when `n_jobs` is greater than logical cores, we get a segfault (when we're lucky). We should just guard that value to be at or less than cores present on the host.

q2_diversity/plugin_setup.py
ebolyen: Jobs is probably better for now, since it sounds like this is actually CPU threads, and not a process thread.

wasade: It's process threads (i.e., `pthread`) that we bind to cores to reduce thread hopping and to leverage NUMA's default local allocation policy.

ElDeveloper: When I wrote this, I knew it was misleading, but carried on since it matches beta_phylogenetic. Would probably be nice to have a standard designation for "workers" all throughout QIIME.

q2_diversity/_beta/_method.py
ElDeveloper: Great catch!!! 👏

q2_diversity/_beta/_method.py
ElDeveloper: I am not sure that this is the reason why we've observed failures, for all the unit tests we compute distances on a 2x2 sample matrix, and those tests pass.

q2_diversity/_beta/_method.py
ElDeveloper: `psutil` to the rescure 💪

q2_diversity/plugin_setup.py
ElDeveloper: Done

q2_diversity/plugin_setup.py
ElDeveloper: Done

ci/recipe/meta.yaml
```diff
@@ -32,6 +32,7 @@ requirements:
     - ipywidgets
     - scikit-bio
     - scikit-learn
+    - psutilt
```
ebolyen: extra `t`

ElDeveloper:
Thanks for fixing @ebolyen!

wasade:
Woooo!!! Thank you guys!

On Sep 26, 2017 3:39 PM, "Evan Bolyen" <notifications@github.com> wrote:

Merged #143 <https://github.com/qiime2/q2-diversity/pull/143>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/qiime2/q2-diversity/pull/143#event-1266737632>, or mute
the thread

PR #156, opened 2017-10-18
Rework beta_correlation into mantel

ElDeveloper:
The method uses two distance matrices instead of a distance matrix and a metadata category.


jairideout:
@ElDeveloper approved over Slack, @ebolyen took a quick gander :ok_hand:

PR #160, opened 2017-10-21
ENH: upgrades core-metrics(-phylogenetic) to pipelines

ebolyen:


q2_diversity/plugin_setup.py
```diff
@@ -240,7 +241,11 @@
         ('unweighted_unifrac_pcoa_results', PCoAResults),
         ('weighted_unifrac_pcoa_results', PCoAResults),
         ('jaccard_pcoa_results', PCoAResults),
-        ('bray_curtis_pcoa_results', PCoAResults)
+        ('bray_curtis_pcoa_results', PCoAResults),
+        ('unweighted_unifrac_pcoa_plot', Visualization),
+        ('weighted_unifrac_pcoa_plot', Visualization),
+        ('jaccard_pcoa_plot', Visualization),
+        ('bray_curtis_pcoa_plot', Visualization),
```
ebolyen: Not sure if I like these names

jairideout: What about using `unweighted_unifrac_emperor`, `weighted_unifrac_emperor`, etc.? That'd match the naming style we're using in the Moving Pictures tutorial:

image

PR #175, opened 2018-01-11
MAINT: update copyright year; travis

ebolyen:


licenses/sklearn.txt
```diff
@@ -1,6 +1,6 @@
 New BSD License
 
-Copyright (c) 2007–2017 The scikit-learn developers.
+Copyright (c) 2007–2017-2018 The scikit-learn developers.
```
thermokarst: `2007–2017-2018`

thermokarst: Probably want to double-check if we need to be updating this file at all.

PR #181, opened 2018-02-01
ENH: Add error bars to alpha-rarefaction plots

### LINKED ISSUE #179: add error bars to alpha rarefaction plots
@jairideout brought this up in #130 - it was planned for the initial release, but didn't make it in. Separating this into it's own issue so it's easier to track on the project board.

thermokarst:
Fixes #179 

---

Questions for reviewer(s):

1) What do we think of this style of error bar? I went with simple, but can make quite a few alternatives happen here.
2) The legend symbol for the error bars is not my favorite --- it seems a bit too narrow (hard to click), but, making it wider makes it even harder to know what it does, IMO. I am open to alternatives.
3) Error bars are currently showing the 9%/91% values from the parametric 7-number summary. Should we be doing something else here?

### DIFF
--- a/q2_diversity/_alpha/_visualizer.py
+++ b/q2_diversity/_alpha/_visualizer.py
@@ -367,6 +367,7 @@ def alpha_rarefaction(output_dir: str, table: biom.Table, max_depth: int,
                        context={'metrics': list(metrics),
                                 'filenames': [quote(f) for f in filenames],
                                 'columns': list(columns),
+                                'steps': steps,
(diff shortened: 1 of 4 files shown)


thermokarst:
Thanks @gregcaporaso! As we discussed offline, I retooled this viz to display box plots in the upper figure, instead of the circle scatter plot symbols that were originally displayed. I also added a new "Help" button with some information about how to use the legend, and how to interpret the line charts and box plots.

PR #188, opened 2018-03-01
MAINT: update unit test error message check

jairideout:
There was a minor change to the error message raised by the framework's Metadata object.


thermokarst:
I am not a smart man.

PR #194, opened 2018-04-19
ENH: Add procrustes analysis plugin

ElDeveloper:
Solves the procrustes problem for two ordination matrices. While we can
support N different matrices as inputs, we can't support varidic
outputs, hence this will be deferred to a later PR when the
functionality is available in the framework.

PR #196, opened 2018-04-20
ENH: added citations

### LINKED ISSUE #193: Add Citations
Should use the new citation API in https://github.com/qiime2/qiime2/issues/387

nbokulich:
fixes #193

### DIFF
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -302,15 +302,7 @@ plugin.pipelines.register_function(
     },
     name='Core diversity metrics (phylogenetic and non-phylogenetic)',
     description="Applies a collection of diversity metrics (both "
-                "phylogenetic and non-phylogenetic) to a feature table.",
-    citations=[
-        citations['lozupone2005unifrac'],
-        citations['lozupone2007quantitative'],
-        citations['faith1992conservation'],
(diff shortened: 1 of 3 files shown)


q2_diversity/plugin_setup.py
```diff
@@ -303,6 +303,14 @@
...
     description="Applies a collection of diversity metrics (both "
                 "phylogenetic and non-phylogenetic) to a feature table.",
+    citations=[
+        citations['lozupone2005unifrac'],
+        citations['lozupone2007quantitative'],
+        citations['faith1992conservation'],
+        citations['vazquez2013emperor'],
+        citations['jaccard1908nouvelles'],
+        citations['pielou1966measurement'],
+        citations['sorensen1948method']]
```
ebolyen: I hate to say this after all your work, but the pipelines won't typically need citations on their own, as the actions they call are tracked in provenance (and so are their citations). Pipelines only need references if their doing something like open-reference.

nbokulich: oh right of course! No problem — the pipelines are the minority, and I just copied/pasted the other citations so it really didn't add work.

thermokarst:
Need to make sure we get a citation for this new action: https://github.com/qiime2/q2-diversity/pull/194

PR #201, opened 2018-06-17
`alpha-group-significance`: add raw data download

nbokulich:
fixes #133

PR #202, opened 2018-06-17
`alpha-correlation`: added raw metadata download

### LINKED ISSUE #132: Export data from alpha-correlation
This came up on [the forum](https://forum.qiime2.org/t/link-to-raw-data-behind-figures/650/4?u=thermokarst).

nbokulich:
fixes #132

### DIFF
--- a/q2_diversity/_alpha/_visualizer.py
+++ b/q2_diversity/_alpha/_visualizer.py
@@ -168,9 +168,10 @@ def alpha_correlation(output_dir: str,
             "execute.")
 
     # save out metadata for download in viz
-    md = pd.concat([alpha_diversity, metadata.to_dataframe()],
-                    axis=1, join='inner')
-    md.to_csv(os.path.join(output_dir, 'metadata.tsv'), sep='\t')
+    alpha_diversity.index.name = 'id'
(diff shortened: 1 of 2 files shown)


q2_diversity/_alpha/_visualizer.py
```diff
@@ -167,6 +167,11 @@ def alpha_correlation(output_dir: str,
             "visualizer requires at least one numeric metadata column to "
             "execute.")
 
+    # save out metadata for download in viz
```
thermokarst: Can we take advantage of the Metadata API for this? See `longitudinal volatility` for an example of [merging](https://github.com/qiime2/q2-longitudinal/blob/master/q2_longitudinal/_longitudinal.py#L189-L193) and [saving](https://github.com/qiime2/q2-longitudinal/blob/master/q2_longitudinal/_longitudinal.py#L244).

nbokulich:
@thermokarst how's that look? I will probably need to do something similar to #203 and #201 before you review those, so just want to make sure.

Note that later in the code metadata is being converted to series to do the same sort of merging I did here.

nbokulich:
LGTM thanks @thermokarst !

Issue #207, opened 2018-06-29
Aitchison distance


mortonjt:
It would be really great if we could include the Aitchison distance.

This is a compositional distance metric that is able to better detect compositional changes.
[Here](http://www.pnas.org/content/113/22/E3130) is an example of applying this distance to detect stress related changes where weighted unifrac completely missed.

PR coming up in the next few hours.

ebolyen:
closed by #208

PR #208, opened 2018-06-29
Aitchison distance

mortonjt:
Adding in Aitchison distance as discussed in #207

### DIFF
--- a/q2_diversity/_beta/_method.py
+++ b/q2_diversity/_beta/_method.py
@@ -20,6 +20,13 @@ from q2_types.tree import NewickFormat
 
 from functools import partial
 
+from skbio.stats.composition import clr
+from scipy.spatial.distance import euclidean
+
+
+def aitchison(x, y):
+    return euclidean(clr(x), clr(y))
+
 
 # We should consider moving these functions to scikit-bio. They're part of
 # the private API here for now.
@@ -39,7 +46,7 @@ def non_phylogenetic_metrics():
             'correlation', 'hamming', 'jaccard', 'chebyshev', 'canberra',
             'braycurtis', 'mahalanobis', 'yule', 'matching', 'dice',
             'kulsinski', 'rogerstanimoto', 'russellrao', 'sokalmichener',
-            'sokalsneath', 'wminkowski'}
+            'sokalsneath', 'wminkowski', 'aitchison'}
 
 
 def all_metrics():
@@ -110,13 +117,20 @@ def beta_phylogenetic_alt(table: BIOMV210Format, phylogeny: NewickFormat,
              variance_adjusted=variance_adjusted, bypass_tips=bypass_tips)
 
 
-def beta(table: biom.Table, metric: str, n_jobs: int=1)-> skbio.DistanceMatrix:
+def beta(table: biom.Table, metric: str,
+         pseudocount: float=1.0, n_jobs: int=1)-> skbio.DistanceMatrix:
     if metric not in non_phylogenetic_metrics():
         raise ValueError("Unknown metric: %s" % metric)
+
+    if metric == 'aitchison':
+        counts = table.matrix_data.toarray().astype(int).T + pseudocount
+        metric = partial(metrics[metric])
+    else:
+        counts = table.matrix_data.toarray().astype(int).T
+
     if table.is_empty():
         raise ValueError("The provided table object is empty")
 
-    counts = table.matrix_data.toarray().astype(int).T
     sample_ids = table.ids(axis='sample')
 
     return skbio.diversity.beta_diversity(
--- a/q2_diversity/_beta/_method.py
+++ b/q2_diversity/_beta/_method.py
@@ -123,13 +123,10 @@ def beta(table: biom.Table, metric: str,
     if not (metric in non_phylogenetic_metrics()):
         raise ValueError("Unknown metric: %s" % metric)
 
+    counts = table.matrix_data.toarray().T
     if metric == 'aitchison':
-        counts = table.matrix_data.toarray().astype(int).T + pseudocount
+        counts += pseudocount
         metric = aitchison
-    else:
-        counts = table.matrix_data.toarray().astype(int).T
-
-    validate = True
 
     if table.is_empty():
         raise ValueError("The provided table object is empty")
@@ -140,7 +137,7 @@ def beta(table: biom.Table, metric: str,
         metric=metric,
         counts=counts,
         ids=sample_ids,
-        validate=validate,
+        validate=True,
         pairwise_func=sklearn.metrics.pairwise_distances,
         n_jobs=n_jobs
     )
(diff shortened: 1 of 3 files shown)


mortonjt:
Ok, there were quite a few annoying type issues that popped up.  Managed to work around this by turning the distance metric into a callable

Raised a couple of issues on skbio
https://github.com/biocore/scikit-bio/issues/1580
https://github.com/biocore/scikit-bio/issues/1581

Note that this is a working solution, but there is much more room for speedups by taking advantage of the inline C code in the euclidean calculation as suggested by @ElDeveloper

mortonjt:
And it looks like this is just a server issue regarding travis

q2_diversity/_beta/_method.py
```diff
@@ -110,19 +113,34 @@ def beta_phylogenetic_alt(table: BIOMV210Format, phylogeny: NewickFormat,
...
+    return euclidean(clr(x), clr(y))
+
+
+def beta(table: biom.Table, metric: str,
+         pseudocount: int=1, n_jobs: int=1)-> skbio.DistanceMatrix:
+
+    if not (metric in non_phylogenetic_metrics()):
         raise ValueError("Unknown metric: %s" % metric)
+
+    if metric == 'aitchison':
```
thermokarst: Style comment/ How about something like this:

```python
counts = table.matrix_data.toarray().astype(int).T
if metric == 'aitchison':
    counts = counts + psuedocount
    metric = aitchison

That code is untested, but it might be easier for new readers of the codebase.

File: q2_diversity/_beta/_method.py [hunk 2]
```diff
@@ -110,19 +113,34 @@ def beta_phylogenetic_alt(table: BIOMV210Format, phylogeny: NewickFormat,
...
+        counts = table.matrix_data.toarray().astype(int).T + pseudocount
+        metric = aitchison
+    else:
+        counts = table.matrix_data.toarray().astype(int).T
+
+    validate = True
```
thermokarst: Maybe this should just be [inlined](https://github.com/qiime2/q2-diversity/pull/208/files#diff-adf90299393c8099efa932cfcca88186R143), below?

q2_diversity/plugin_setup.py
```diff
@@ -146,6 +147,8 @@
     },
     parameter_descriptions={
         'metric': 'The beta diversity metric to be computed.',
+        'pseudocount': ('A pseudocount to handle zeros '
```
thermokarst: Maybe we should indicate that this value is ignored in other cases:

> A pseudocount to handle zeros when computing compositional metrics, otherwise, this parameter is ignored.

q2_diversity/plugin_setup.py
```diff
@@ -138,6 +138,7 @@
     function=q2_diversity.beta,
     inputs={'table': FeatureTable[Frequency]},
     parameters={'metric': Str % Choices(beta.non_phylogenetic_metrics()),
+                'pseudocount': Int,
```
thermokarst: Should a range be applied to this? Setting a pseudocount of 0 seems silly. As well, would there be a use case for negative pseudocounts? If not, you could do something like this:

```python
from qiime2.plugin import Range

...

'pseudocount': Int % Range(1, None),
```

q2_diversity/_beta/_method.py
thermokarst: As well, if we are introducing new pseudocount functionality, a test or two should probably be added to cover this case.

q2_diversity/plugin_setup.py
mortonjt: 👍 Although in future cases we will want to handle scenarios to have fractional pseudocounts, which is why I'm pushing in a PR into skbio to relax the type checks

q2_diversity/_beta/_method.py
```diff
@@ -110,19 +113,31 @@ def beta_phylogenetic_alt(table: BIOMV210Format, phylogeny: NewickFormat,
              variance_adjusted=variance_adjusted, bypass_tips=bypass_tips)
 
 
-def beta(table: biom.Table, metric: str, n_jobs: int=1)-> skbio.DistanceMatrix:
-    if metric not in non_phylogenetic_metrics():
+def aitchison(x, y, **kwds):
+    return euclidean(clr(x), clr(y))
```
ElDeveloper: I know this is related to #213 but does the same problem arise if you compute the CLR transform for all the samples in the feature table and then individually call euclidean? i.e. if you change this function to:

```python
def aitchison(x, y, **kwds):
    return euclidean(x, y)
```

mortonjt:
Yeah... That's that I tried after you suggested it - there were weird
issues popping up with the type errors. But maybe we can read dress with
the new skbio release

mortonjt:
@thermokarst , I believe that addresses the outstanding issues.  Thanks!

thermokarst:
Thanks @mortonjt --- looks like the unit tests are failing on travis - want to take a look at that before I review again?

mortonjt:
Yeah, I was a little confused about those errors... Let me take another look

mortonjt:
@thermokarst on other thoughts, do you think we can wait for the next version of skbio to be released?  We just got the relaxed type checking pushed in [here](https://github.com/biocore/scikit-bio/pull/1583) and are aiming to get the next release out by the end of July.

thermokarst:
@mortonjt - sure! Ping us here when you want to put it back on the radar. Thanks!

mortonjt:
It looks like scikit-bio just got a release: https://github.com/biocore/scikit-bio/pull/1591

Will push out changes later this week.

mortonjt:
@ElDeveloper comments are addressed

ElDeveloper:
Looks good to me @mortonjt! Note that we'll likely need https://github.com/qiime2/q2-types/pull/191 to be merged first.

mortonjt:
I think the tests should now pass - @thermokarst could we restart the travis tests?

thermokarst:
@mortonjt it won't make a difference until the changes are reflected here: https://github.com/qiime2/environment-files

Busywork is still cooking away (it takes 2+ hours for it to do its thing) 🔥

mortonjt:
Ok!

thermokarst:
Hey @mortonjt - reran the tests on travis, there are still failures. Let me know if you need a hand looking at this. Thanks!

mortonjt:
Looks like the type checks are performing as expected -- the negative values from the clr are breaking the checks.

I don't think we can get the optimized aitchison distance in this round -- the type checks are still to stringent.  Sorry @ElDeveloper @wasade

mortonjt:
@thermokarst tests are passing -- what do you think?

mortonjt:
Thanks!!

mortonjt:
Just realized that we should probably make the pseudocount more compatible with the functionality here: https://github.com/qiime2/q2-gneiss/pull/43

Not a big deal though -- should be an easy fix for the next release.

https://github.com/qiime2/q2-diversity/blob/master/q2_diversity/_beta/_method.py#L117

ebolyen:
Yeah that's fine. I think we're looking at really breaking about `beta` and `alpha` soon into something more manageable. This is good evidence for how the composability kind of breaks when you mash every metric ever into one super method.

PR #217, opened 2018-08-20
[work in progress]: Update ordination class to support new scikit-bio pcoa API

HannesHolste:
Update ordination class to support new scikit-bio's fast heuristic pcoa method https://github.com/biocore/scikit-bio/pull/1569

Changes are fully backwards-compatible.

TO-DO:
1. Wait until skbio PR is merged and scikit-bio 0.5.4 is released.
2. Update qiime2 dependencies to use scikit-bio-0.5.4
3. Test and merge this PR


HannesHolste:
I anticipate all the TODOs to be complete today. Currently preparing skbio for a new release with Yoshiki.

HannesHolste:
@thermokarst just wanted to put this on your radar to mark for review the QIIME2 release.

HannesHolste:
@thermokarst How to trigger q2d2 to generate a new dev/staging environment file based on the latest package dependencies in q2-types (which contains skbio 0.5.4, which is needed by this PR)? https://github.com/qiime2/environment-files/tree/master/latest/staging

q2_diversity/_ordination.py
```diff
@@ -9,5 +9,10 @@
 import skbio.stats.ordination
 
 
-def pcoa(distance_matrix: skbio.DistanceMatrix) -> skbio.OrdinationResults:
-    return skbio.stats.ordination.pcoa(distance_matrix)
+def pcoa(distance_matrix: skbio.DistanceMatrix, method: str = 'eigh',
+         number_of_dimensions: int = 0,
+         inplace: bool = False) -> skbio.OrdinationResults:
```
ebolyen: Is `inplace` necessary at this time? From a qiime2 perspective, it's always parsing a file, in the future we might do something clever like skipping the format/disk if actions are pipelined together, but in that case we'd need to _not_ mutate at all.

Otherwise, the only time this would come up is if you were using the function directly, in which case, you're much better of using `skbio.stats.ordination.pcoa` which intends to be a library function.

ebolyen:
It's in progress already, https://busywork.qiime2.org. In a few hours there will be a new environment file for travis.

q2_diversity/tests/test_ordination.py
```diff
@@ -22,5 +23,40 @@ def test_pcoa(self):
...
+        assert_ordination_results_equal(results, results_inplace,
+                                        ignore_directionality=True,
+                                        ignore_method_names=True)
+
+        # Test invalid parameters for number_of_dimensions
+        with self.assertRaises(ValueError):
+            pcoa(dm1, method='fsvd', number_of_dimensions=-1)
+
+        with self.assertRaises(ValueError):
+            pcoa(dm1, method='fsvd', number_of_dimensions=-100)
```
ebolyen: These tests aren't necessary, the range predicate precludes number_of_dimensions being negative.

q2_diversity/plugin_setup.py
```diff
@@ -205,7 +205,11 @@
 plugin.methods.register_function(
     function=q2_diversity.pcoa,
     inputs={'distance_matrix': DistanceMatrix},
-    parameters={},
+    parameters={
+        'method': Str,
+        'number_of_dimensions': Int % Range(0, None),
+        'inplace': Bool
+    },
```
ebolyen: Add corresponding `parameter_descriptions` for these attributes. You'll definitely want to explain the behavior of 0 for `number_of_dimensions`

q2_diversity/_ordination.py
ebolyen: Yep, I think that's a good plan, let's leave inplace=False for QIIME 2 :+1:

q2_diversity/_ordination.py
```diff
@@ -9,5 +9,10 @@
 import skbio.stats.ordination
 
 
-def pcoa(distance_matrix: skbio.DistanceMatrix) -> skbio.OrdinationResults:
-    return skbio.stats.ordination.pcoa(distance_matrix)
+def pcoa(distance_matrix: skbio.DistanceMatrix, method: str = 'eigh',
```
ElDeveloper: Being that this is a high-level interface (unlike scikit-bio), I would suggest that the function signature be simplified to:

```python
def pcoa(distance_matrix: skbio.DistanceMatrix, number_of_dimensions: int=None) -> skbio.OrdinationResults
    if number_of_dimensions is None:
        # calculate full decomposition using eigh
        # using inplace=False
    else:
        #  calculate the decomposition only for the `number_of_dimensions` using fsvd
        # using inplace=True

    return ordination      
```

The logic being that it doesn't make a lot of sense to select a subset of dimensions for a full decomposition and that when that is done, we might as well just avoid the extra work.

Sorry if I'm coming in too late to the discussion.

ebolyen: I like this even better!

q2_diversity/_ordination.py
```diff
@@ -10,8 +10,20 @@
 import pandas as pd
 
 
-def pcoa(distance_matrix: skbio.DistanceMatrix) -> skbio.OrdinationResults:
-    return skbio.stats.ordination.pcoa(distance_matrix)
+def pcoa(distance_matrix: skbio.DistanceMatrix,
+         number_of_dimensions: int = 0) -> skbio.OrdinationResults:
```
ebolyen: I did like @ElDeveloper's initial idea of using `None` instead of `0` here. Since it would mean *any* value passed to `number_of_dimensions` results in fsvd. Then the range can be `Range(1, None)`.

ElDeveloper: I would be OK with that since there are no contracts or requirements that there's a 1:1 match between interfaces.

q2_diversity/_ordination.py
```diff
@@ -10,8 +10,19 @@
 import pandas as pd
 
 
-def pcoa(distance_matrix: skbio.DistanceMatrix) -> skbio.OrdinationResults:
-    return skbio.stats.ordination.pcoa(distance_matrix)
+def pcoa(distance_matrix: skbio.DistanceMatrix,
+         number_of_dimensions: int = None) -> skbio.OrdinationResults:
+    if number_of_dimensions == None:
```
ebolyen: Linter wants this to be `is None`

q2_diversity/plugin_setup.py
```diff
@@ -205,13 +202,39 @@
...
                             'computed.')
     },
-    parameter_descriptions={},
+    parameter_descriptions={
+        'method': "Eigendecomposition method to use in performing PCoA. "
+                  "By default, uses SciPy's eigh, which computes exact "
+                  "eigenvectors and eigenvalues for all dimensions. The "
+                  "alternate method, fsvd, uses faster heuristic "
+                  "eigendecomposition but loses accuracy. The magnitude "
+                  "of accuracy lost is dependent on dataset.",
```
ebolyen: Drop this description since it doesn't exist anymore.

q2_diversity/_ordination.py
```diff
@@ -10,8 +10,19 @@
...
+        # calculate full decomposition using eigh
+        return skbio.stats.ordination.pcoa(distance_matrix, method='eigh',
+                                           inplace=False)
+    else:
+        # calculate the decomposition only for the `number_of_dimensions`
+        # using fast heuristic eigendecomposition (fsvd)
+        return skbio.stats.ordination.pcoa(
+            distance_matrix, method='fsvd',
+            number_of_dimensions=number_of_dimensions,
+            inplace=False)
```
ElDeveloper: If using FSVD I think the default should be `inplace=True` since the approximation also makes an implication about the size of the dataset and we know (from experience) that inplace=True can make or break your computation.

HannesHolste:
@ElDeveloper thanks, implemented! @ebolyen travis is passing!

ElDeveloper:
Great, thanks!

PR #218, opened 2018-08-22
ENH: Add pcoa biplot method

ElDeveloper:
Allows users to project descriptors into an existing PCoA matrix

Brief summary of the Pull Request, including any issues it may fix using the GitHub closing syntax:

Adds a new method to project variables into a PCoA matrix. This depends on busywork to update the conda recipe (as the method depends on scikit-bio 0.5.3).

A subsequent q2-emperor PR will allow for users to see these results.

**questions**:

I wondered whether or not I should add an annotation to the PCoAResults type. Namely something like `PCoAResults[Biplot]`. This way this method would output a `PCoAResults[Biplot]` artifact that can only be consumed by `q2-emperor`'s `biplot` method. If reviewers think this makes sense, I'll add the proper changes in q2-types.

### DIFF
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -231,10 +231,10 @@ plugin.methods.register_function(
     output_descriptions={'biplot': 'The resulting PCoA matrix.'},
     name='Principal Coordinate Analysis Biplot',
     description="Project features into a principal coordinates matrix. The "
-                "features used can be environmental variables or the features"
-                " used to compute the distance matrix. It is recommended that"
-                " these variables be normalized in cases of dimensionally "
-                "heterogeneous physical variables.",
+                "features used should be the features used to compute the "
+                "distance matrix. It is recommended that these variables be"
+                " normalized in cases of dimensionally heterogeneous physical"
+                " variables.",
     citations=[citations['legendrelegendre']]
 )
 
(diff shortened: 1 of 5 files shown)


ebolyen:
I think using `PCoAResults % Properties("Biplot")` would make sense, since the biplot is just extra information making the PCoAResults a more refined type. It would be usable anywhere `PCoAResults` is useable, but then methods can request the specialized variant if needed explicitly.

q2_diversity/plugin_setup.py
```diff
@@ -217,6 +217,27 @@
...
+    outputs=[('biplot', PCoAResults % Properties('biplot'))],
+    input_descriptions={
+        'pcoa': 'The PCoA where the features will be projected onto.',
+        'features': 'Variables to project onto the PCoA matrix'
+    },
+    parameter_descriptions={},
+    output_descriptions={'biplot': 'The resulting PCoA matrix.'},
+    name='Principal Coordinate Analysis Biplot',
+    description="Project features into a principal coordinates matrix. The "
+                "features used can be environmental variables or the features"
```
ebolyen: Should the description skip the environmental variables, or should we add support for it in this action?

It does seem that the API get's a little tricky since metadata would be a parameter. (would both become optional and mutually exclusive?)

ebolyen:
Just one comment, otherwise LGTM

q2_diversity/plugin_setup.py
ElDeveloper: Indeed it gets tricky. I've opted to remove that for now. Later, when the API has been exercised a bit more, we can discuss of better ways to do this.

PR #219, opened 2018-08-22
ENH: Add permdisp as a group significance method

ElDeveloper:
Adds permdisp to the beta-group-significance method.


ElDeveloper:
🎉

PR #232, opened 2019-01-02
DOC: add FSVD citation; note Legendre and Legendre for PCoA citation

wasade:
Adds FSVD citation, and also updates the pcoa action to cite Legendre and Legendre (which I believe is correct...).


ElDeveloper:
I believe the original reference is Gower 1966: http://hsta559s12.pbworks.com/w/file/fetch/53338762/gower1966.pdf

Though it is true that the underlying implementation is entirely based on the outline from L&L.

PR #238, opened 2019-02-06
ENH: batch alpha calculations to reduce table / tree size

wasade:
The alpha calculation code casts the input feature table into a dense representation. For large studies, this represents a large memory requirement. This memory requirement is further increased for Faith's PD as, with the present implementation, an intermediary counts array is produced that is N * M where N is the number of samples and M is the number of vertices in the tree.

This code automatically partitions the input feature table into smaller sizes, and then aggregates the resulting alpha diversity scores into a single pandas `Series`. 

I opted to not expose the `batch_size` parameter as 1000 is pretty large, but small enough that even for a table with 1M features, the memory footprint is reasonably small. The downside to a smaller batch size is it increases the number of executions of `alpha_diversity` and the associated validation overhead.

### DIFF
--- a/q2_diversity/_alpha/_method.py
+++ b/q2_diversity/_alpha/_method.py
@@ -48,7 +48,6 @@ def alpha_phylogenetic(table: biom.Table, phylogeny: skbio.TreeNode,
 
     results = []
     for counts, sample_ids, feature_ids in _batch_table(table):
-
         try:
             result = skbio.diversity.alpha_diversity(metric=metric,
                                                      counts=counts,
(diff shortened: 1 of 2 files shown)


thermokarst:
@wasade - @ChrisKeefe is currently in the process of factoring out the alpha commands into a new util plugin - https://github.com/qiime2/q2-diversity-lib. Maybe it makes sense to close this PR and reopen on that new plugin, once https://github.com/qiime2/q2-diversity-lib/pull/2 is merged. What do you think?

wasade:
The delta here is pretty small, maybe just bring over these changes into q2-diversity-lib?

thermokarst:
Yep, I think that makes sense to me!

wasade:
Wonderful, thanks! Should I close this PR then?

PR #239, opened 2019-02-13
BUG: adonis, normalize MD header

thermokarst:
This could also be `'sample-id'` - thoughts?


ElDeveloper:
+1, although it is now solved, in older versions of pandas, there was a 
bug with the index name setters; this lead to a similar unit test being 
added to emperor.

PR #251, opened 2019-07-10
Alpha phylogenetic alt for stacked faith

gwarmstrong:
This PR adds the necessary code to use the implementation of Faith's PD in https://github.com/biocore/unifrac . This method is recommended for large datasets.

Since it is fairly new, it is provided as an alternative: `qiime diversity alpha-phylogenetic-alt`

I have included some plots benchmarking the `Stacked Faith` implementation from UniFrac against `scikit-bio`, which is currently used in `qiime diversity alpha-phylogenetic`

time_memory_combined

PR #252, opened 2019-07-10
up unifrac version to expose high performance faith's pd

gwarmstrong:
This PR ups the required version of `unifrac` to expose Faith's PD via `unifrac` in order to trigger QIIME2 continuous integration prior to `q2_alpha_phylogenetic_alt` functionality can be properly tested in #251

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -34,7 +34,7 @@ requirements:
     - psutil
     - natsort
     - biom-format >=2.1.5,<2.2.0
-    - unifrac >=0.9.2,<0.10.0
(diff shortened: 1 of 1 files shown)


ebolyen:
Does the bioconda build that's already up work for this? Or should we wait for the current PR that pins some dependencies more specifically?

gwarmstrong:
> Does the bioconda build that's already up work for this? Or should we wait for the current PR that pins some dependencies more specifically?

@ebolyen I think the current bioconda build should work!

ebolyen:
Then let's take advantage of busywork being green to get this through!

PR #263, opened 2019-10-07
FIX: problem with wrong encoded filenames, closes #197

### LINKED ISSUE #197: "No such file" when viewing beta-group-significance visualization
A visualization created with "qiime diversity beta-group-significance" and ~opened via q2view~ viewed using any available method will not load the images created. From @thermokarst, looks like a bug related to something in the metadata column names not being escaped correctly --- if you export the viz all the graphs are there, they just arent loading as expected in q2view

angrybee:
This fixed my problem with misencoded filenames in `qiime diversity beta-group-significance` , closes #197

I had problems with groupnames like _NucleoSpin® Soil - Macherey Nagel_.
The file is saved as _NucleoSpin%C2%AE+Soil+-+Macherey+Nagel-boxplots.pdf_ and linked in the HTML-file as _NucleoSpin®+Soil+-+Macherey+Nagel-boxplots.pdf_.

Removing the encoding parts worked. Tested with FF 69.0.1 (64-bit), Chromium 77.0.3865.90.

### DIFF
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -191,9 +191,9 @@ def beta_group_significance(output_dir: str,
         plt.tight_layout()
         fig = ax.get_figure()
         fig.savefig(os.path.join(output_dir, '%s-boxplots.png' %
-                                 urllib.parse.quote_plus(str(group_id))))
+                                 str(group_id)))
         fig.savefig(os.path.join(output_dir, '%s-boxplots.pdf' %
-                                 urllib.parse.quote_plus(str(group_id))))
+                                 str(group_id)))
         fig.clear()
 
     pairs_summary.to_csv(os.path.join(output_dir, 'raw_data.tsv'), sep='\t')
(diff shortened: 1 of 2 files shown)


ebolyen:
Thanks @angrybee! I don't think the approach here will be generally filesystem safe (which is the reason it gets encoded, say you had a group named `A/B`), but clearly what we have isn't quite working either.

I kind of suspect the issue may lie with the template. Is there a better way for us to "urlencode" in the template? I suspect that the `replace(' ', '+')` is insufficient, hence failing on `®`

angrybee:
Hi @ebolyen,

Using the urlencode also does not work. If the HTML-links have the correct name with the encoded filename the browser decodes it and cannot find the files which were saved with the encoded name.

In the current version I escape the filename/group_id. Nearly everything which is not alphanummeric gets substitute by an underscore. If anyone misses a valid character in the future it should be easy to modify the exceptions in the implemented function.
Is this a workaround we can live with?

ebolyen:
closing in favor of #264 

Thanks for putting in the work for trying to fix this, we really appreciate it! Your comment helped me figure out what was happening behind the scenes here.

PR #264, opened 2019-10-30
BUG: fix escaping of filenames in beta-group-sig

ebolyen:
It turns out this does kind of makes sense, it's just super weird.

Many thanks to @angrybee for the [initial legwork here](https://github.com/qiime2/q2-diversity/pull/263#issuecomment-539444693).


ebolyen:
fixes #197

ebolyen:
Take 2 [can be viewed here](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum.qiime2.org%2Fuploads%2Fshort-url%2FgL1z6rMa3m4XAsrnG2ncN1rAtEv.qzv).

This also makes sense, as the `+` gets encoded as `%2B` the second time around rather than staying `+`. Instead, I'm just doing no-shortcut quoting, which seems to work better.

PR #268, opened 2020-02-13
Pandas 1 fixes

andrewsanchez:


q2_diversity/_alpha/_visualizer.py
```diff
@@ -241,7 +241,10 @@ def _reindex_with_metadata(column, columns, merged):
     merged.sort_index(axis=0, ascending=True, inplace=True)
     merged = merged.groupby(level=[column])
     counts = merged.count()
-    counts.drop(columns, axis=1, inplace=True, level=0)
+    # Removes the column name used to set the index of `merged` above
+    col_diff = set(columns) - set([column])
+    if col_diff:
+        counts.drop(col_diff, axis=1, inplace=True, level=0)
```
andrewsanchez: The first kwarg taken by ` DataFrame.drop` is `labels` and it accepts a single label or list-like.  Passing the set directly seems to work find here, but we can make it a `pd.Index` or list if that is preferable.

q2_diversity/_alpha/_visualizer.py
```diff
@@ -237,12 +237,15 @@ def alpha_correlation(output_dir: str,
 
 
 def _reindex_with_metadata(column, columns, merged):
-    merged.set_index(column, inplace=True)
-    merged.sort_index(axis=0, ascending=True, inplace=True)
-    merged = merged.groupby(level=[column])
-    counts = merged.count()
-    counts.drop(columns, axis=1, inplace=True, level=0)
-    median_ = merged.median()
+    reindexed = merged.set_index(column)
```
andrewsanchez: Avoid mutating `merged` by assigning to a new DataFrame `reindexed`.

q2_diversity/_alpha/_visualizer.py
nbokulich: nah, this looks good to me

q2_diversity/_alpha/_visualizer.py
```diff
@@ -237,12 +237,15 @@ def alpha_correlation(output_dir: str,
...
+    reindexed.sort_index(axis=0, ascending=True, inplace=True)
+    grouped = reindexed.groupby(level=[column])
+    counts = grouped.count()
+    # Removes the column name used to set the index of `merged` above
+    col_diff = set(columns) - set([column])
+    if col_diff:
```
nbokulich: will col_diff ever be false (empty)? I am trying to wrap my head around this one... 

If yes, could you point to the unit test that covers this case? (or create such a unit test)

andrewsanchez: Yes.  For example, in `test_alpha_rarefaction.py:test_alpha_rarefaction_with_filtered_metadata_columns`, `col_diff` will be empty.  There are several test cases where `_reindex_with_metadata` is called with arguments such that `[column] == columns`, e.g., `column='pet'` and `columns=['pet']`, which will result in `col_diff` being empty.

nbokulich:
Thanks for clarifying... LGTM

PR #271, opened 2020-04-18
Added jensenshannon metric

patrickimran:
This PR just adds another non-phylogenetic metric, Jensen Shannon distance, to the set of non_phylogenetic_metrics. It is one of those additional metrics implemented in [scipy.spatial.distance.pdist](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html) which can be passed to skbio.diversity.

### DIFF
--- a/q2_diversity/_beta/_method.py
+++ b/q2_diversity/_beta/_method.py
@@ -111,7 +111,7 @@ def beta(table: biom.Table, metric: str,
 
         return (1. / nnz) * np.sum(np.abs(x_ - y_) / (x_ + y_))
 
-    def jensenshannon(x, y, **kwds):
+    def jensen_shannon(x, y, **kwds):
         return jensenshannon(x, y)
 
     if metric == 'aitchison':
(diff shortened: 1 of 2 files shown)


thermokarst:
Thanks @patrickimran! @ChrisKeefe is currently working on moving most (all?) of these metrics out into a new standalone plugin (https://github.com/qiime2/q2-diversity-lib). @ChrisKeefe - how do you want to address integrating the changes proposed in this PR?

ChrisKeefe:
Thanks for the contribution, @patrickimran ! I don't think `q2-diversity-lib` is _quite_ ready for outside contributions yet, but :crossed_fingers: that should change in the next month or two. I think it's best to bring this into `q2-diversity` for now, and I can port it over to `diversity-lib` later. Or you can, if that would bring you joy :wink: 

If you're interested in contributing diversity measures to `q2-diversity-lib` in the future, we would _love_ the support. `div-lib` allows us to provide measures as individual Actions, with more robust testing, more flexible semantic typing, individualized citations and documentation, and other niceties. Let me know. :frog:

PR #272, opened 2020-05-11
Transition alpha_phylogenetic_alt to alpha_phylogenetic

gwarmstrong:
This PR is to transition the `alpha_phylogenetic_alt` method to `alpha_phylogenetic` as the main method for computing Faith's PD with `q2-diversity`. As a reminder, the implementation for the current `alpha_phylogenetic_alt` yields speed improvements and a memory reduction of several orders of magnitude. See [here](https://github.com/biocore/unifrac/pull/65) for more details.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -34,7 +34,7 @@ requirements:
     - psutil
     - natsort
     - biom-format >=2.1.5,<2.2.0
-    - unifrac >=0.9.2,<0.10.0
+    - unifrac >=0.10.0
     - qiime2 {{ release }}.*
     - q2templates {{ release }}.*
     - q2-types {{ release }}.*
--- a/q2_diversity/__init__.py
+++ b/q2_diversity/__init__.py
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-from ._alpha import (alpha, alpha_phylogenetic, alpha_phylogenetic_alt,
+from ._alpha import (alpha, alpha_phylogenetic,
                      alpha_group_significance, alpha_correlation,
                      alpha_rarefaction)
 from ._beta import (beta, beta_phylogenetic, bioenv,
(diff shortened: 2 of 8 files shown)


thermokarst:
Hey @gwarmstrong, this'll be great to get into the next release. Looks like the test suite is failing - I'll leave this for you to investigate, but, if you need a hand just ping us here. Thanks!

gwarmstrong:
@thermokarst this fell off the radar for a little bit. Thanks for the ping!

gwarmstrong:
So one of the issues with my addition is that `alpha_rarefaction` takes a `biom.Table` and the alternative`alpha_phylogenetic` takes a `BIOMV210Format`. I think the appropriate band-aid is to transform the table into the appropriate type right before plugging it into `alpha_phylogenetic`.

Can you recommend the best way to access the `biom.Table` -> `BIOMV210Format` transformer?

gwarmstrong:
It looks like I may be able to use this?
`from qiime2.plugin.util import transform`

thermokarst:
> It looks like I may be able to use this?
> `from qiime2.plugin.util import transform`

Yes indeed! That util is the best way to manually invoke a transformation.

gwarmstrong:
By the way, this is passing the test suite now. It is just failing on coverage, since there were already a few uncovered lines in beta and I decreased the total line count.

thermokarst:
Sweet, thanks! Coverage measurements can be so silly sometimes - I'll pop this over into the "Review" column - thanks!

gwarmstrong:
@thermokarst your changes look good to me! Preserving the old behavior makes sense.

I also made changes to `alpha_rarefaction` [here](https://github.com/qiime2/q2-diversity/pull/272/files#diff-938bddf064046a59925f75268cfd23b7R298-R367), should we add an `alpha_rarefaction_old` to mirror `alpha_phylogenetic_old`?

Thanks!

thermokarst:
> should we add an alpha_rarefaction_old to mirror alpha_phylogenetic_old

Hmm... I think I'm fine leaving out any kind of escape hatch for the alpha rarefaction viz, since it is diagnostic in nature.

PR #278, opened 2020-06-16
BUG: Corrects typo in mantel skbio URL

ChrisKeefe:
Bad URL was causing 404 errors. This PR removes duplicated `generated` to align with correct address.

### DIFF
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -655,7 +655,7 @@ plugin.visualizers.register_function(
                 'the same; the distance matrices will be reordered before '
                 'applying the Mantel test.\n\nSee the scikit-bio docs for '
                 'more details about the Mantel test:\n\n'
-                'http://scikit-bio.org/docs/latest/generated/generated/'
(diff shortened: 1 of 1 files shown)


q2_diversity/plugin_setup.py
```diff
@@ -655,7 +655,7 @@
                 'the same; the distance matrices will be reordered before '
                 'applying the Mantel test.\n\nSee the scikit-bio docs for '
                 'more details about the Mantel test:\n\n'
-                'http://scikit-bio.org/docs/latest/generated/generated/'
+                'http://scikit-bio.org/docs/latest/generated/'
                 'skbio.stats.distance.mantel.html',
```
andrewsanchez, line 659: What do you think about dropping the `.html` from the end of the link?  The CLI chops the link up across two lines, making it unclickable, at least for me in iTerm2:

image

http://scikit-bio.org/docs/latest/generated/skbio.stats.distance.mantel still works and fits on a single line.

```suggestion
                'http://scikit-bio.org/docs/latest/generated/'
                'skbio.stats.distance.mantel',
```

Issue #255, opened 2019-07-11
coverage < 100% in q2_diversity/_beta/_method.py


gwarmstrong:
**Improvement Description**
There are two lines that are not covered in `q2_diversity/_beta/_method.py`, which means anytime the number of lines in this repo is decreased, it will cause the `coveralls` to fail (e.g., #254 ).

The culprits are:

https://github.com/qiime2/q2-diversity/blob/4c235a003b6ab8ad6cca78935c33365a8903f1af/q2_diversity/_beta/_method.py#L73-L74

and

https://github.com/qiime2/q2-diversity/blob/4c235a003b6ab8ad6cca78935c33365a8903f1af/q2_diversity/_beta/_method.py#L120-L122

The first can probably be tested with `unittest.mock` and the second can be testing with something like [this](http://code.activestate.com/recipes/580716-unit-testing-nested-functions)

**Current Behavior**
Unittests fail if total number of lines is decreased in `q2_diversity/_beta/_method.py`

**Proposed Behavior**
Unittests pass if total number of lines is decreased in `q2_diversity/_beta/_method.py`

**Questions**
* I can move forward and make the changes, the `psutils` case should be relatively easy to cover, but it might require an annoying number of additional lines to test the negative values in `canberra_adkins`. Additionally, as it stands I do not think this error can be invoked since `skbio.diversity.beta_diversity` already checks for negative values. Would it make sense to remove the error check in the `canberra_adkins` function? Since this check is already being made somewhere.

ChrisKeefe:
#281 replaces `_method.py` with `_pipeline.py`. 
https://github.com/qiime2/q2-diversity-lib/pull/6/files# resolves the first issue.
https://github.com/qiime2/q2-diversity-lib/pull/24 should resolve the canberra-adkins issue by removing that guard. (as @gwarmstrong suggested, we'll trust `skbio` on this)

PR #281, opened 2020-06-25
ENH: Route all diversity calculations through diversity-lib

### LINKED ISSUE #280: ENH: Route all alpha and beta diversity calculations through q2-diversity-lib
**Improvement Description**
The following Actions should be made into Pipelines, dependent on Actions from `q2-diversity-lib`:
- `alpha`
- `alpha phylogenetic`
- `beta`
- `beta phylogenetic`

`q2-diversity-lib`, in turn, will dispatch these calls to the best available implementation of the requested method. See https://github.com/qiime2/q2-diversity-lib/issues/22

**Current Behavior**
`core-metrics` and `core-metrics-phylogenetic` call methods from `q2-diversity-lib` for alpha and beta diversity calculations. 
The `alpha` and `beta` Actions, however, are called directly from `q2-diversity`. 

This proposal makes `q2-diversity-lib` uniquely responsible for determining what implementation of a given method should be used. This allows `q2-diversity` to stop caring about what methods have been implemented/constrained in `q2-diversity-lib` at any given moment, and which methods will simply be passed through to an appropriate dependency.

**Proposed Behavior**
- The four actions above are re-implemented as Pipelines
- https://github.com/qiime2/q2-diversity-lib/issues/22 has been closed, and dispatch functions are properly meeting the needs of this issue
- Any other impacted methods (e.g. `alpha-rarefaction` are cleaned up as needed)

**Comments**
1. This will likely require minor changes to other Actions that use these functions - e.g. `alpha_phylogenetic`

### LINKED ISSUE #284: MAINT: Remove deprecated `alpha-phylogenetic-` methods
**Improvement Description**
`alpha-phylogenetic-old` and `alpha-phylogenetic-alt` are currently deprecated and should be removed in `2020.8`

- [x] Remove methods
- [x] Remove associated tests
- [x] Cleanup

ChrisKeefe:
Closes #280.
See also qiime2/q2-diversity-lib#21

Closes #284.

~~Note: Renames "weighted_unifrac" to "weighted_unnormalized_unifrac", aligning `q2-diversity` with `q2-diversity-lib` and the underlying `unifrac` package. This is a breaking API change.~~


ChrisKeefe:
This tests clean locally, but tests will continue to fail here until https://github.com/qiime2/q2-diversity-lib/pull/21 is merged.

ChrisKeefe:
This PR is ready for a _preliminary_ review - does anything here need major course correction? 

Only `beta_phylogenetic` has been converted into a pipeline at this time. I'll continue transitioning other Actions on a separate branch while this is in review, so this doesn't become a moving target.

ChrisKeefe:
After beginning work on dispatch for nonphylogenetic beta diversity measures, I've decided to move dispatch from `diversity-lib` to `diversity`, preventing some gross circular dependency action. Removed this from In Review until that's wrapped up.

thermokarst:
# Executive Summary

Thanks @ChrisKeefe, these q2-diversity/q2-diversity-lib changes are looking great! After spending some time going through these PRs, I have a handful of thoughts, which I will try to outline here.

TLDR: I think there is at least one too many levels of indirection here, and with a little restructuring this code can be simplified conceptually and structurally.

## Overall structuring

The role of the q2-diversity-lib plugin is to serve as library code for q2-diversity. The q2-diversity-lib code could just as easily live inside of q2-diversity itself, but we chose to factor it out as a plugin for a few reasons, one of them being just for the sake of giving you an opportunity to make a plugin from scratch.  With that in mind, I think it is important to emphasize that the use of this q2-diversity-lib library code does not need to be minimized in q2-diversity, in fact, I think that minimizing/masking q2-diversity-lib invocations in q2-diversity is problematic.

If we emphasize that the actions available in q2-diversity-lib are meant to compose together into bigger pipelines in q2-diversity, we get the best of both worlds - compartmentalization of code (in q2-diversity-lib) and easy to follow pipelines (in q2-diversity).

With that in mind, generally speaking:

- Individual methods belong in q2-diversity-lib
- Pipelines belong in q2-diversity
- Visualizers that aren't able to consume Pipelines via the QIIME 2 framework
  should consume the pipeline functions via plain python

## q2-diversity-lib

### Mappings

The proliferation of data-lookups-as-utility-functions-but-only-sometimes is a bit confusing:

- `alpha.implemented_phylogenetic_measures_dict()`
- `alpha.implemented_phylogenetic_measures()`
- `alpha.all_phylogenetic_measures_alpha` (note not a function)
- `alpha.implemented_nonphylogenetic_measures_dict()`
- `alpha.implemented_nonphylogenetic_measures()`
- `alpha.unimplemented_nonphylogenetic_measures()`
- `alpha.all_nonphylogenetic_measures_alpha()`
- `beta.implemented_nonphylogenetic_metrics_dict()`
- `beta.implemented_nonphylogenetic_metrics()`
- `beta.unimplemented_nonphylogenetic_metrics()`
- `beta.all_nonphylogenetic_measures_beta()`
- `beta.implemented_phylogenetic_metrics_dict()`
- `beta.unimplemented_phylogenetic_metrics_dict()`
- `beta.all_phylogenetic_measures_dict()`
- `beta.all_phylogenetic_measures_beta()`

Instead of this, I propose one coherent data structure that encodes all of this information in one pass:

```python
MAPPINGS = {
    'PHYLO': {
        'IMPL': {'faith_pd'},
        'UNIMPL': set(),
    },
    'NONPHYLO': {
        'IMPL': {'observed_features', 'pielou_e', 'shannon'},
        'UNIMPL': {'ace', 'chao1', 'chao1_ci', 'berger_parker_d',
                   'brillouin_d', 'dominance', 'doubles', 'enspie', 'esty_ci',
                   'fisher_alpha', 'goods_coverage', 'heip_e',
                   'kempton_taylor_q', 'margalef', 'mcintosh_d', 'mcintosh_e',
                   'menhinick', 'michaelis_menten_fit', 'osd', 'robbins',
                   'simpson', 'simpson_e', 'singles', 'strong', 'gini_index',
                   'lladser_pe', 'lladser_ci'},
    },
}
```

This lends itself well to "natural reading", especially when chaining together:

```python
metrics = MAPPINGS['NONPHYLO']['IMPL'] | MAPPINGS['NONPHYLO']['UNIMPL']
implemented_metrics = MAPPINGS['NONPHYLO']['IMPL']
```

Also, this has the advantage of only exporting one object, rather than a half dozen loosely related ones.

You also might've noticed that there are dictionaries in the tips, I don't think you'll need to include those dicts, and hopefully you'll see why, elsewhere in this review. Right now I am envisioning one `MAPPINGS` in the `alpha` module, and one in `beta`.

Also, I'm open to other names (`METRICS`, for example), and different layouts of the data, this is just to get you started with some ideas.

#### Translations

I like the idea of including translations, and I mention it here as a subsection of mappings, but only because I see it as a loosely related idea. I think you will be able to move any translation functionality into the q2-diversity Pipelines, discussed later in this review.

### Methods

I think the following QIIME 2 Methods should be exposed in this plugin (will reference the path to the view function, for clarity):

- `alpha.faith_pd`
- `alpha.observed_features`
- `alpha.pielou_evenness`
- `alpha.shannon_entropy`
- `alpha.alpha_passthrough`
- `beta.bray_curtis`
- `beta.jaccard`
- `beta.unweighted_unifrac`
- `beta.weighted_unifrac`
- `beta.beta_passthrough`
- `beta.beta_phylogenetic_passthrough`

You'll notice there is no `alpha_phylogenetic_passthrough` Method, only because there are no unimplemented alpha phylogenetic metrics in q2-diversity.

#### `alpha_passthrough`

This method will handle just the `MAPPINGS['NONPHYLO']['UNIMPL']` metrics, and might look something like this:

```python
@_disallow_empty_tables
def alpha_passthrough(table: biom.Table, metric: str) -> pd.Series:
    counts = table.matrix_data.toarray().astype(int).T
    sample_ids = table.ids(axis='sample')

    result = skbio.diversity.alpha_diversity(metric=metric, counts=counts,
                                             ids=sample_ids)
    result.name = metric
    return result
```

#### `beta_passthrough`

This method will handle just the `MAPPINGS['NONPHYLO']['UNIMPL']` metrics.

Same idea as `alpha_passthrough` above - defer to skbio, don't worry about fancy signatures, citations, etc. Basically just moving the `beta` method over from q2-diversity.

#### `beta_phylogenetic_passthrough`

This method will handle just the `MAPPINGS['PHYLO']['UNIMPL']` metrics.

Same idea as `alpha_passthrough` above - defer to skbio, don't worry about fancy signatures, citations, etc. Basically just moving the `beta_phylogenetic` method over from q2-diversity.

### Pipelines

There will be none in this plugin. That means _removing_:

- `alpha.alpha_dispatch`
- `alpha.alpha_phylogenetic_dispatch`
- `beta.beta_dispatch`
- `beta.beta_phylogenetic_dispatch`

The bodies of these pipelines will be moved to q2-diversity.

### Pipelines as Python Functions

There will be none in this plugin. That means _removing_:

- `alpha.alpha_rarefaction_dispatch`
- `alpha.alpha_rarefaction_phylogenetic_dispatch`
- `beta.skbio_dispatch`
- `beta.unifrac_beta_dispatch`

### Visualizers

There will be none in this plugin.

## q2-diversity

### Methods

Several existing methods will be converted to Pipelines (see below).

The following Methods will be _removed_:

- `_alpha.alpha`
- `_alpha.alpha_phylogenetic`
- `_alpha.alpha_phylogenetic_alt`
- `_alpha.alpha_phylogenetic_old`
- `_beta.beta`
- `_beta.beta_phylogenetic`

### Pipelines

New pipelines (former Methods):

- `_alpha.alpha`
- `_alpha.alpha_phylogenetic`
- `_beta.beta`
- `_beta.beta_phylogenetic`

These pipelines can all basically have the same implementations as the `*_dispatch` pipelines that you added to q2-diversity-lib.

#### `alpha`

This is more-or-less a copy-paste from your current `alpha_dispatch` pipeline:

```python
def alpha(ctx, table, metric, drop_undefined_samples=None):
    metrics = MAPPINGS['NONPHYLO']['IMPL'] | MAPPINGS['NONPHYLO']['UNIMPL']
    implemented_metrics = MAPPINGS['NONPHYLO']['IMPL']

    if metric not in metrics:
        raise ValueError("Unknown metric: %s" % metric)

    metric = ALPHA_TRANSLATION[metric]

    if metric in implemented_metrics:
        func = ctx.get_action('diversity_lib', metric)
        if 'drop_undefined_samples' in signature(func).parameters:
            func = partial(func, table=table,
                           drop_undefined_samples=drop_undefined_samples)
        else:
            if drop_undefined_samples:
                warnings.warn(f"The {metric} metric does not support dropping "
                              "undefined samples.")
            func = partial(func, table=table)
    else:
        func = ctx.get_action('diversity_lib', 'alpha_passthrough')
        func = partial(func, table=table, metric=metric)

    result = func()
    return tuple(result)
```

### Pipelines as Python Functions

The following Pipelines shall be used as QIIME 2 Pipelines, as well as vanilla Python code:

- `_alpha.alpha`
- `_alpha.alpha_phylogenetic`
- `_beta.beta`
- `_beta.beta_phylogenetic`

This can be handled by building a dependency-injection mock, to fill in the `ctx` argument:

```python
class PyContext:
    def __init__(self):
        self.mappings = {
            'diversity_lib': {
                'faith_pd': q2_diversity_lib.alpha.faith_pd,
                'observed_features': q2_diversity_lib.alpha.observed_features,
                'pielou_e': q2_diversity_lib.alpha.pielou_evenness,
                'shannon': q2_diversity_lib.alpha.shannon_entropy,
                'alpha_passthrough': q2_diversity_lib.alpha.alpha_passthrough,
                'bray_curtis': q2_diversity_lib.beta.bray_curtis,
                'jaccard': q2_diversity_lib.beta.jaccard,
                # ... etc etc etc
            },
        }

    def get_action(self, plugin, action):
        return self.mappings[plugin][action]
```

### Visualizers

The following visualizers will need to be updated to invoke the "Pipelines as Python Functions" listed above:

- `_alpha.alpha_rarefaction`
- `_beta.beta_rarefaction`

Using the `PyContext` class above, we can "inject" the vanilla Python functions in:

```python
...

ctx = PyContext()

for d, i in itertools.product(depth_range, iter_range):
    rt = rarefy(feature_table, d)
    for m in metrics:
        if m in phylogenetic_metrics():
            rt_p = transform(rt, to_type=BIOMV210Format,
                             from_type=biom.Table)
            vector = alpha_phylogenetic(
                    ctx=ctx, table=rt_p, metric=m, phylogeny=phylogeny)
        else:
            vector = alpha(ctx=ctx, table=rt, metric=m,
                           drop_undefined_samples=False)
        data[m][(d, i)] = vector
return data

...
```

## Summary and Conclusion

Like I said above, this all looks great, and I think the main parts of the implementation are there, but I think we can do a lot to help streamline it and reduce the complexity of the Pipeline -> Pipeline -> Method workflow that is currently proposed. Let me know if you have any questions or comments. I'm sure I missed one or two things in here, but I have played with all of these proposed changes locally and can confirm that the plan proposed here works.

🦖

q2_diversity/tests/test_alpha_rarefaction.py
```diff
@@ -174,7 +174,7 @@ def test_invalid_invocations(self):
                 alpha_rarefaction(output_dir, t, max_depth=200,
                                   metadata=md, metrics=set(['faith_pd']))
 
-            with self.assertRaisesRegex(ValueError, 'Unknown metric: pole'):
+            with self.assertRaisesRegex(TypeError, 'pole.*incompatible'):
                 alpha_rarefaction(output_dir, t, max_depth=200,
                                   metadata=md, metrics=set(['pole-position']))
```
thermokarst, line 179: image

ChrisKeefe: Yeah, I was glad I got to leave that in. Especially after I had to cut `unwigged_unifunc` along with `diversity_lib.test_get_nonexistent_action` this afternoon.

q2_diversity/_alpha/_visualizer.py
```diff
@@ -16,20 +16,16 @@
...
 import numpy as np
 import pandas as pd
-import qiime2
 from statsmodels.sandbox.stats.multicomp import multipletests
-import q2templates
 import biom
 import itertools
+
+import qiime2
+import q2templates
```
thermokarst, line 24: These two can be put back where they came from - the pep8 convention is to group external deps, these were already fine as-is.

q2_diversity/_alpha/_visualizer.py
```diff
@@ -16,20 +16,16 @@
...
+from q2_diversity import _alpha
```
thermokarst, line 25: This is one step away from creating a circular import nightmare, how about

```suggestion
from .pipeline import (...bleh)
```

q2_diversity/_alpha/_visualizer.py
```diff
@@ -291,20 +287,26 @@ def _compute_rarefaction_data(feature_table, min_depth, max_depth, steps,
...
-            else:
-                vector = alpha(table=rt, metric=m)
-            data[m][(d, i)] = vector
-    return data
+    if phylogeny:
+        phylogeny = qiime2.Artifact.import_data('Phylogeny[Rooted]', phylogeny)
+
+    with qiime2.sdk.Context() as scope:
+        for depth, i in itertools.product(depth_range, iter_range):
+            rt = rarefy(feature_table, depth)
```
thermokarst, line 295: This should be called through the `ctx.get_action`, too.

q2_diversity/_alpha/_visualizer.py
```diff
@@ -291,20 +287,26 @@ def _compute_rarefaction_data(feature_table, min_depth, max_depth, steps,
...
+                if metric in _alpha.all_phylo_metrics:
+                    alpha_phylo = scope.ctx.get_action('diversity',
+                                                       'alpha_phylogenetic')
+                    vector = alpha_phylo(table=rt, metric=metric,
+                                         phylogeny=phylogeny)
+                else:
+                    alpha = scope.ctx.get_action('diversity', 'alpha')
+                    vector = alpha(table=rt, metric=metric)
+                vector = vector[0].view(pd.Series)
+                data[metric][(depth, i)] = vector
```
thermokarst, line 308:
```suggestion
            for metric in metrics:
                if metric in _alpha.all_phylo_metrics:
                    alpha_phylo = scope.ctx.get_action('diversity',
                                                       'alpha_phylogenetic')
                    vector, = alpha_phylo(table=rt, metric=metric,
                                         phylogeny=phylogeny)
                else:
                    alpha = scope.ctx.get_action('diversity', 'alpha')
                    vector, = alpha(table=rt, metric=metric)
                vector = vector.view(pd.Series)
                data[metric][(depth, i)] = vector
```

q2_diversity/_beta/_pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+
+def all_metrics():
+    return all_phylo_metrics | all_nonphylo_metrics
+
+
+def beta_phylogenetic(ctx, table, phylogeny,
+                      metric, threads=1,
+                      variance_adjusted=False,
+                      alpha=None,
+                      bypass_tips=False):
```
thermokarst, line 26:
```suggestion
def beta_phylogenetic(ctx,
                      table,
                      phylogeny,
                      metric,
                      threads=1,
                      variance_adjusted=False,
                      alpha=None,
                      bypass_tips=False):
```

might as well get everyone on their own lines, at this point.

q2_diversity/_beta/_pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+    if metric not in all_phylo_metrics:
+        raise ValueError("Unknown metric: %s" % metric)
```
thermokarst, line 28: Let's drop this guard - QIIME 2 protects us from this case, and we don't support calling via python, directly. These "duplicated guards" are leftovers from our early days with developing QIIME 2 plugins.

q2_diversity/_beta/_pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if alpha is not None and metric != 'generalized_unifrac':
+        raise ValueError('The alpha parameter is only allowed when the choice'
+                         ' of metric is generalized_unifrac')
```
thermokarst, line 32: Isn't this already implemented in q2-div-lib? Why duplicate it here?

q2_diversity/_beta/_pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+            and not variance_adjusted:
+        func = ctx.get_action('diversity_lib', metric_tr)
+        result = func(table, phylogeny, threads=threads,
+                      bypass_tips=bypass_tips)
+    else:
+        # handle unimplemented unifracs
+        func = ctx.get_action('diversity_lib', 'beta_phylogenetic_passthrough')
+        result = func(table, phylogeny, metric=metric_tr, threads=threads,
+                      variance_adjusted=variance_adjusted, alpha=alpha,
+                      bypass_tips=bypass_tips)
```
thermokarst, line 47: I really think this would be easier to reason about without the `translate_metric_name`, and instead just covering all of the metrics in the `METRICS` constant.

```suggestion
    metric = METRICS['NAME_TRANSLATION'][metric]

    # HACK: this logic will be simpler once the remaining unifracs are done
    if metric in ('unweighted_unifrac', 'weighted_unifrac') \
            and not variance_adjusted:
        func = ctx.get_action('diversity_lib', metric)
        result = func(table, phylogeny, threads=threads,
                      bypass_tips=bypass_tips)
    else:
        # handle unimplemented unifracs
        func = ctx.get_action('diversity_lib', 'beta_phylogenetic_passthrough')
        result = func(table, phylogeny, metric=metric, threads=threads,
                      variance_adjusted=variance_adjusted, alpha=alpha,
                      bypass_tips=bypass_tips)
```

q2_diversity/_alpha/_pipeline.py
```diff
@@ -0,0 +1,43 @@
...
+        raise ValueError("Unknown metric: %s" % metric)
+
+    metric_tr = translate_metric_name(metric, metric_name_translations)
+
+    if metric in implemented_metrics:
+        func = ctx.get_action('diversity_lib', metric_tr)
+        result = func(table=table)
+    else:
+        func = ctx.get_action('diversity_lib', 'alpha_passthrough')
+        result = func(table=table, metric=metric_tr)
```
thermokarst, line 41:
```suggestion
    metric = METRICS['NAME_TRANSLATIONS'][metric]
    if metric in METRICS['NONPHYLO']['IMPL']:
        func = ctx.get_action('diversity_lib', metric)
        result = func(table=table)
    else:
        func = ctx.get_action('diversity_lib', 'alpha_passthrough')
        result = func(table=table, metric=metric)
```

q2_diversity/_alpha/_visualizer.py
```diff
@@ -291,20 +287,26 @@ def _compute_rarefaction_data(feature_table, min_depth, max_depth, steps,
...
-                rt_p = transform(rt, to_type=BIOMV210Format,
-                                 from_type=biom.Table)
-                vector = alpha_phylogenetic(table=rt_p, metric=m,
-                                            phylogeny=phylogeny)
-            else:
-                vector = alpha(table=rt, metric=m)
-            data[m][(d, i)] = vector
-    return data
+    if phylogeny:
+        phylogeny = qiime2.Artifact.import_data('Phylogeny[Rooted]', phylogeny)
```
thermokarst, line 291: If we move this into the `scope` context manager, then you can use `make_artifact`, which tracks the lifetime of the artifact for you:

https://github.com/qiime2/qiime2/blob/55f122c689d647743c207d9891adcbf1def40897/qiime2/sdk/context.py#L41-L52

q2_diversity/tests/test_alpha.py
```diff
@@ -19,189 +19,121 @@
...
-            alpha_phylogenetic_old(
-                table=t, phylogeny=tree, metric='not-a-metric')
+    def test_alpha_phylo_metric(self):
+        with self.assertRaisesRegex(TypeError, 'faith_pd.*incompatible'):
+            self.alpha(table=self.t, metric='faith_pd')
+        # Check vanilla python function as well
+        with qiime2.sdk.Context() as scope:
+            with self.assertRaisesRegex(ValueError,
+                                        'Unknown metr.*not-a-metric'):
+                alpha(ctx=scope.ctx, table=self.t, metric='not-a-metric')
```
thermokarst, line 80: The second half of this test is unnecessary, and I think it should be removed:

```suggestion
    def test_alpha_phylo_metric(self):
        with self.assertRaisesRegex(TypeError, 'faith_pd.*incompatible'):
            self.alpha(table=self.t, metric='faith_pd')
```

thermokarst: Please remove any other similar "vanilla python" checks, as well.

q2_diversity/_alpha/_visualizer.py
ChrisKeefe: Thanks for bringing this up. It's something I've been wondering about lately. The PEP8 language is:

>     Standard library imports.
>     Related third party imports.
>     Local application/library specific imports.

For a platform built across many repos, where some are truly third-party and some are not, it almost feels like there's another class of imports: platform-local but non-package-local (i.e. packages native to the QIIME 2 platform, especially if built by core QIIME 2 devs). These land in a gap between "third party imports" and "local application" imports. 

I'm over-engineering here, and will just put "non-package-local" imports in with the other third-party tools we use in future.

q2_diversity/_beta/_pipeline.py
ChrisKeefe: Removed from `alpha`, `alpha_phylo`, `beta`, `beta_phylo`.

q2_diversity/tests/test_alpha.py
ChrisKeefe: These were added as much to fix test coverage (#255), as they were to support vanilla python imports. Now that the 'invalid metric' guards have been removed from the alpha and beta dispatch functions, these have been removed without negatively impacting test coverage.

q2_diversity/_beta/_pipeline.py
ChrisKeefe: This ensures that if users ask `q2_diversity.beta_phylogenetic` to call `weighted_unifrac` or `unweighted_unifrac` with an `alpha`, they are told not to do so. Without this seemingly duplicate guard, the `alpha` argument is silently dropped during dispatch, and the user gets no feedback that they're doing something wrong. I don't like the extra guard, but I like the alternatives less.

Options:
- Duplicate the guard
- User gets no feedback about their error if using `weighted_` or `unweighted_unifrac` with an `alpha`
- We remove the guard from `div-lib` in favor of `diversity.beta_phylogenetic`. This leaves users of `div-lib.beta_phylogenetic_passthrough` with no feedback about their error. Only an improvement because few people will use that directly.
- We add an irrelevant `alpha` param to `q2_diversity_lib.unweighted_unifrac` and `q2_diversity_lib.weighted_unifrac`, for the sole purpose of guarding against this behavior only in `div-lib`. Just no.

I got the whiteboard out for this one, but I may still have missed a better solution. Let me know if you have a better insight.

q2_diversity/_alpha/_visualizer.py
ChrisKeefe: resolved in 171f005

q2_diversity/_alpha/_visualizer.py
ChrisKeefe: resolved in 960a7f0

q2_diversity/_beta/_pipeline.py
thermokarst: Let's leave as is (duplicate the guard). I added TODOs to both plugins to drop when we can handle optional type-mapped params.

thermokarst:
Going to merge now - travis is failing because the staging env is missing the latest div-lib. I tested locally and all seems in order, no need to wait 2-3 hrs for the new env.

PR #282, opened 2020-07-10
ENH: adonis: validate metadata to check for NaNs

### LINKED ISSUE #265: ENH: adonis: validate metadata to check for NaNs
**Improvement Description**
Generate a user-friendly error when NaNs are present in the metadata

**Current Behavior**
R spews some nonsense about non-conformable arrays

**Proposed Behavior**
Catch this with some simple metadata validation, or capture the error message and report something more user-friendly.

collin-le:
Fixes #265

### DIFF
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -365,6 +365,8 @@ def adonis(output_dir: str,
         for i in t.factors:
             metadata.get_column(i.name())
 
+    # Check metadata for NaNs
+
     # Run adonis
     results_fp = os.path.join(output_dir, 'adonis.tsv')
     with tempfile.TemporaryDirectory() as temp_dir_name:
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -369,6 +369,7 @@ def adonis(output_dir: str,
     metadata = metadata.to_dataframe()
     if metadata.isnull().values.any() == True:
         raise ValueError("Adonis cannot run with NaNs in metadata.")
+    metadata = qiime2.Metadata(metadata)
 
     # Run adonis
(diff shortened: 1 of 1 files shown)


q2_diversity/_beta/_visualizer.py
```diff
@@ -365,6 +365,12 @@ def adonis(output_dir: str,
         for i in t.factors:
             metadata.get_column(i.name())
 
+    # Check metadata for NaNs
+    metadata = metadata.to_dataframe()
+    if metadata.isnull().values.any() == True:
+        raise ValueError("Adonis cannot run with NaNs in metadata.")
+    metadata = qiime2.Metadata(metadata)
```
thermokarst, line 372:
```suggestion
    metadata_df = metadata.to_dataframe()
    if metadata_df.isnull().values.any():
        raise ValueError("Adonis cannot run with NaNs in metadata.")
```

collin-le: Yes, your suggested changes look great! I noticed the adonis test file, but was not sure if it was necessary/what would need to be done for that. Can you please explain what needs to be done to create the test?

thermokarst: Thanks @collin-le!

I think in the adonis test suite we could just copy a test like this:

https://github.com/qiime2/q2-diversity/blob/406fedf67a574ecd8885f63dc58e06bd3ff5147a/q2_diversity/tests/test_adonis.py#L78-L84

and change the test data to be something with NaNs in the metadata. Then, the `self.assertRaisesRegex` part could match on some part of your error message above (`"Adonis cannot run with NaNs in metadata."`).

thermokarst:
@collin-le   - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

PR #292, opened 2020-08-20
Exposing unifrac.meta

wasade:
This pull request is dependent on qiime2/q2-diversity-lib#26. 

The known TODO items are:

- [x] adjust plugin_setup text to better reflect the method (currently effectively a copy/paste of `beta_phylogenetic`)
- [x] add support for `meta` specific arguments
- [x] augment testing to run through meta

### DIFF
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -111,8 +111,7 @@ plugin.pipelines.register_function(
                           RelativeFrequency |
                           PresenceAbsence]],
             'phylogeny': List[Phylogeny[Rooted]]},
-    parameters={'metric': Str % Choices(beta.METRICS['PHYLO']['IMPL'] |
-                                        beta.METRICS['PHYLO']['UNIMPL']),
+    parameters={'metric': Str % Choices(beta.METRICS['PHYLO']['UNIMPL']),
                 'threads': Int % Range(1, None) | Str % Choices(['auto']),
                 'variance_adjusted': Bool,
                 'alpha': Float % Range(0, 1, inclusive_end=True),
@@ -155,7 +154,7 @@ plugin.pipelines.register_function(
                          'aggregated'
     },
     output_descriptions={'distance_matrix': 'The resulting distance matrix.'},
-    name='Beta diversity (phylogenetic)',
+    name='Beta diversity (phylogenetic) over multiple trees and tables',
     description="Computes a distance matrix for all pairs of samples in the "
                 "set of feature table and phylogeny pairs, using the unifrac "
                 "implementation of the selected beta diversity metric.",
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -8,7 +8,7 @@
 
 from qiime2.plugin import (Plugin, Str, Properties, Choices, Int, Bool, Range,
                            Float, Set, Visualization, Metadata, MetadataColumn,
-                           Categorical, Numeric, Citations)
+                           Categorical, Numeric, Citations, List)
 
 import q2_diversity
 from q2_diversity import _alpha as alpha
@@ -102,6 +102,55 @@ plugin.pipelines.register_function(
     description=("Computes a user-specified phylogenetic beta diversity metric"
                  " for all pairs of samples in a feature table.")
 )
+
+
+plugin.pipelines.register_function(
+    function=q2_diversity.beta_phylogenetic_meta,
+    inputs={'table': List[FeatureTable[Frequency |
+                          RelativeFrequency |
+                          PresenceAbsence]],
+            'phylogeny': List[Phylogeny[Rooted]]},
+    parameters={'metric': Str % Choices(beta.METRICS['PHYLO']['IMPL'] |
+                                        beta.METRICS['PHYLO']['UNIMPL']),
+                'threads': Int % Range(1, None) | Str % Choices(['auto']),
+                'variance_adjusted': Bool,
+                'alpha': Float % Range(0, 1, inclusive_end=True),
+                'bypass_tips': Bool},
+    outputs=[('distance_matrix', DistanceMatrix)],
+    input_descriptions={
+        'table': ('The feature table containing the samples over which beta '
+                  'diversity should be computed.'),
+        'phylogeny': ('Phylogenetic tree containing tip identifiers that '
+                      'correspond to the feature identifiers in the table. '
+                      'This tree can contain tip ids that are not present in '
+                      'the table, but all feature ids in the table must be '
+                      'present in this tree.')
+    },
+    parameter_descriptions={
+        'metric': 'The beta diversity metric to be computed.',
+        'threads': threads_description,
+        'variance_adjusted': ('Perform variance adjustment based on Chang et '
+                              'al. BMC Bioinformatics 2011. Weights distances '
+                              'based on the proportion of the relative '
+                              'abundance represented between the samples at a'
+                              ' given node under evaluation.'),
+        'alpha': ('This parameter is only used when the choice of metric is '
+                  'generalized_unifrac. The value of alpha controls importance'
+                  ' of sample proportions. 1.0 is weighted normalized UniFrac.'
+                  ' 0.0 is close to unweighted UniFrac, but only if the sample'
+                  ' proportions are dichotomized.'),
+        'bypass_tips': ('In a bifurcating tree, the tips make up about 50% of '
+                        'the nodes in a tree. By ignoring them, specificity '
+                        'can be traded for reduced compute time. This has the'
+                        ' effect of collapsing the phylogeny, and is analogous'
+                        ' (in concept) to moving from 99% to 97% OTUs')
+    },
+    output_descriptions={'distance_matrix': 'The resulting distance matrix.'},
+    name='Beta diversity (phylogenetic)',
+    description=("Computes a user-specified phylogenetic beta diversity metric"
+                 " for all pairs of samples in the feature tables.")
+)
+
 
 plugin.pipelines.register_function(
     function=q2_diversity.beta,
--- a/q2_diversity/plugin_setup.py
+++ b/q2_diversity/plugin_setup.py
@@ -19,6 +19,7 @@ from q2_types.distance_matrix import DistanceMatrix
 from q2_types.sample_data import AlphaDiversity, SampleData
 from q2_types.tree import Phylogeny, Rooted
 from q2_types.ordination import PCoAResults
+from unifrac._meta import CONSOLIDATIONS
 
 citations = Citations.load('citations.bib', package='q2_diversity')
 
@@ -115,16 +116,18 @@ plugin.pipelines.register_function(
                 'threads': Int % Range(1, None) | Str % Choices(['auto']),
                 'variance_adjusted': Bool,
                 'alpha': Float % Range(0, 1, inclusive_end=True),
-                'bypass_tips': Bool},
+                'bypass_tips': Bool,
+                'weights': List[Float],
+                'consolidation': Str % Choices(list(CONSOLIDATIONS))},
     outputs=[('distance_matrix', DistanceMatrix)],
     input_descriptions={
-        'table': ('The feature table containing the samples over which beta '
-                  'diversity should be computed.'),
-        'phylogeny': ('Phylogenetic tree containing tip identifiers that '
-                      'correspond to the feature identifiers in the table. '
-                      'This tree can contain tip ids that are not present in '
-                      'the table, but all feature ids in the table must be '
-                      'present in this tree.')
+        'table': 'The feature tables containing the samples over which beta '
+                 'diversity should be computed.',
+        'phylogeny': 'Phylogenetic trees containing tip identifiers that '
+                     'correspond to the feature identifiers in the table. '
+                     'This tree can contain tip ids that are not present in '
+                     'the table, but all feature ids in the table must be '
+                     'present in this tree.'
     },
     parameter_descriptions={
         'metric': 'The beta diversity metric to be computed.',
@@ -143,12 +146,19 @@ plugin.pipelines.register_function(
                         'the nodes in a tree. By ignoring them, specificity '
                         'can be traded for reduced compute time. This has the'
                         ' effect of collapsing the phylogeny, and is analogous'
-                        ' (in concept) to moving from 99% to 97% OTUs')
+                        ' (in concept) to moving from 99% to 97% OTUs'),
+        'weights': 'The weight applied to each tree/table pair. This tuple is '
+                   'expected to be in index order with tables and phylogenies.'
+                   ' Default is to weight each tree/table pair evenly.',
+        'consolidation': 'The matrix consolidation method, which determines '
+                         'how the individual distance matrices are '
+                         'aggregated'
     },
     output_descriptions={'distance_matrix': 'The resulting distance matrix.'},
     name='Beta diversity (phylogenetic)',
-    description=("Computes a user-specified phylogenetic beta diversity metric"
-                 " for all pairs of samples in the feature tables.")
+    description="Computes a distance matrix for all pairs of samples in the "
+                "set of feature table and phylogeny pairs, using the unifrac "
+                "implementation of the selected beta diversity metric.",
 )
 
 
--- a/q2_diversity/tests/test_beta.py
+++ b/q2_diversity/tests/test_beta.py
@@ -119,26 +119,27 @@ class BetaDiversityTests(TestPluginBase):
         for method, table_, tree_ in ((self.beta_phylogenetic, t, tree),
                                       (self.beta_phylogenetic_meta, [t, t],
                                        [tree, tree])):
-            actual = method(
-                table=table_, phylogeny=tree_, metric='unweighted_unifrac')
+            with self.subTest(method=method):
+                actual = method(
+                    table=table_, phylogeny=tree_, metric='unweighted_unifrac')
 
-            self.assertEqual(len(actual), 1)
+                self.assertEqual(len(actual), 1)
 
-            self.assertEqual(repr(actual.distance_matrix.type),
-                             'DistanceMatrix')
+                self.assertEqual(repr(actual.distance_matrix.type),
+                                 'DistanceMatrix')
 
-            # expected computed with skbio.diversity.beta_diversity
-            expected = skbio.DistanceMatrix([[0.00, 0.25, 0.25],
-                                             [0.25, 0.00, 0.00],
-                                             [0.25, 0.00, 0.00]],
-                                            ids=['S1', 'S2', 'S3'])
+                # expected computed with skbio.diversity.beta_diversity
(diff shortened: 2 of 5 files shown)


wasade:
Brief note, I'm refactoring some of the `beta_phylogenetic` tests so they can also use this method. Hoping to push next commit tomorrow

wasade:
Expecting to push a commit or two today

wasade:
Specifying an alpha for generalized unifrac, to use with `meta`, is not currently supported due to biocore/unifrac#117. This is at the moment reflected in the test code, though it may be necessary to do something in the library. 

This is a bug but I'm not sure if I can do a minor unifrac release quickly enough for this. Possibly... the bug is simple.

wasade:
Just had another reasonable item pop up on unifrac for some performance related commits that aren't in conda, so will try to get that release out today

wasade:
@thermokarst, I *think* the build is not getting qiime2/q2-diversity-lib#26? I may be looking at the failures incorrectly though

thermokarst:
> @thermokarst, I _think_ the build is not getting [qiime2/q2-diversity-lib#26](https://github.com/qiime2/q2-diversity-lib/pull/26)?

That is correct! BW is churning away now, hopefully we'll have new env files in the next few hrs. Worth noting, DockerHub's new rate limiting scheme is putting a damper on BW right now, so we might have some CI downtime while that is addressed. 🤞

wasade:
Rate limits were meant to be broken...? 🚀

q2_diversity/plugin_setup.py
```diff
@@ -103,6 +104,64 @@
...
 
+
+plugin.pipelines.register_function(
+    function=q2_diversity.beta_phylogenetic_meta,
+    inputs={'table': List[FeatureTable[Frequency |
+                          RelativeFrequency |
+                          PresenceAbsence]],
+            'phylogeny': List[Phylogeny[Rooted]]},
+    parameters={'metric': Str % Choices(beta.METRICS['PHYLO']['IMPL'] |
+                                        beta.METRICS['PHYLO']['UNIMPL']),
```
thermokarst, line 115: The current implementation of `beta_phylogenetic_meta` in this PR doesn't actually have any discretely implement individual metrics ([for example](https://github.com/qiime2/q2-diversity-lib/blob/dead1b4ad531d19fd06516b844c20a59d9803495/q2_diversity_lib/beta.py#L182-L201)), if that is the intent, please remove the `IMPL` metrics:

```suggestion
    parameters={'metric': Str % Choices(beta.METRICS['PHYLO']['UNIMPL']),
```

q2_diversity/plugin_setup.py
```diff
@@ -103,6 +104,64 @@
...
+                        ' (in concept) to moving from 99% to 97% OTUs'),
+        'weights': 'The weight applied to each tree/table pair. This tuple is '
+                   'expected to be in index order with tables and phylogenies.'
+                   ' Default is to weight each tree/table pair evenly.',
+        'consolidation': 'The matrix consolidation method, which determines '
+                         'how the individual distance matrices are '
+                         'aggregated'
+    },
+    output_descriptions={'distance_matrix': 'The resulting distance matrix.'},
+    name='Beta diversity (phylogenetic)',
```
thermokarst, line 158: Should we say something about the "meta"ness? Otherwise this is identical to `beta-phylogenetic`'s description:

Screen Shot 2020-11-19 at 2 16 27 PM

q2_diversity/tests/test_beta.py
```diff
@@ -114,24 +115,30 @@ def test_beta_empty_table(self):
     def test_beta_phylogenetic(self):
         t = self.two_feature_table
         tree = self.three_feature_tree
-        actual = self.beta_phylogenetic(
-            table=t, phylogeny=tree, metric='unweighted_unifrac')
 
-        self.assertEqual(len(actual), 1)
+        for method, table_, tree_ in ((self.beta_phylogenetic, t, tree),
+                                      (self.beta_phylogenetic_meta, [t, t],
+                                       [tree, tree])):
```
thermokarst, line 121: Commenting once, but this is applicable to all of these new implementation loops.

Perhaps we should use [`unittest.subTest`](https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.subTest) to differentiate the tests in the runner report?

```suggestion
        for method, table_, tree_ in ((self.beta_phylogenetic, t, tree),
                                      (self.beta_phylogenetic_meta, [t, t],
                                       [tree, tree])):
                with self.subTest(method=method):
                     ...
```

q2_diversity/tests/test_beta.py
```diff
@@ -153,230 +163,310 @@ def test_beta_phylogenetic_unknown_metric(self):
...
+        table = self.crawford_table
+        tree = self.crawford_tree
+
+        for method, table_, tree_ in ((self.beta_phylogenetic, table, tree),
+                                      (self.beta_phylogenetic_meta,
+                                       [table, table],
+                                       [tree, tree])):
+            with self.assertRaises(ValueError):
+                # cannot guarantee that this will always be true, but it would
+                # be odd to see a machine with these many CPUs
```
thermokarst, line 466: Quantum computing, man.

thermokarst (changes requested):

q2_diversity/plugin_setup.py
wasade: Not fully sure I follow. Unweighted, weighted, variance adjusted and generalized should be suitable for meta unifrac, and all of these are implemented in `unifrac`? I may be misunderstanding something here

q2_diversity/plugin_setup.py
wasade: :+1: good point :)

q2_diversity/tests/test_beta.py
wasade: Did not know about `subTest`, that is super cool.

q2_diversity/tests/test_beta.py
wasade: I think that came from a copypasta...

q2_diversity/plugin_setup.py
thermokarst: The reason we set up q2-diversity-lib is so that we can implement per-metric-qiime2-methods (like what I linked to above). This is great for provenance etc. The `IMPL` metrics are the list of metrics that have their own q2-diversity-lib "methods". The `UNIMPL` are metrics that don't have their own individual div-lib methods. If you peek at `beta` you'll see this illustrated more clearly:

https://github.com/qiime2/q2-diversity/blob/60de8d541e724dae7c453736599de78e6c1891fe/q2_diversity/_beta/_pipeline.py#L45-L52

ChrisKeefe:
Super excited to see this getting added. Just wondering whether there's a reason we went with `table` and `phylogeny` instead of `tables` and `phylogenies` if users are likely to pass `List`s of them. Same question applies to https://github.com/qiime2/q2-diversity-lib/pull/26/

q2_diversity/plugin_setup.py
```diff
@@ -103,6 +104,63 @@
                  " for all pairs of samples in a feature table.")
 )
 
+
+plugin.pipelines.register_function(
+    function=q2_diversity.beta_phylogenetic_meta,
+    inputs={'table': List[FeatureTable[Frequency |
+                          RelativeFrequency |
+                          PresenceAbsence]],
```
ebolyen, line 112: Quick question, can the list be polymorphic as described, or should it be monomorphic like `List[FeatureTable[Frequency] | FeatureTable[RelativeFrequency] | FeatureTable[PresenceAbsence]]` ?

It seems like this would depend on if it was weighted or not? We could constrain the input with a typemap if so:

```pytb
In [1]: from qiime2.plugin import TypeMap, Str, Choices, Visualization, List

In [2]: from q2_types.feature_table import FeatureTable, Frequency, RelativeFrequency, PresenceAbsence

In [3]: T_table, P_metric, _ = TypeMap({
   ...:     (List[FeatureTable[Frequency]
   ...:          | FeatureTable[RelativeFrequency]
   ...:          | FeatureTable[PresenceAbsence]],
   ...:      Str % Choices("unweighted_unifrac")): Visualization,  # a convention for no-op w.r.t. output
   ...:     (List[FeatureTable[Frequency] | FeatureTable[RelativeFrequency]],
   ...:      Str % Choices('weighted_unifrac', 'weighted_normalized_unifrac', 'generalized_unifrac')): Visualization
   ...: })

In [4]: T_table
Out[4]: List[FeatureTable[Frequency] | FeatureTable[RelativeFrequency] | FeatureTable[PresenceAbsence]]¹ | List[FeatureTable[Frequency] | FeatureTable[RelativeFrequency]]²

In [5]: P_metric
Out[5]: Str % Choices('unweighted_unifrac')¹ | Str % Choices('weighted_unifrac', 'weighted_normalized_unifrac', 'generalized_unifrac')²
```

Although... to be honest I kind of hate what that type will look like in our interfaces, so I am good ignoring this...

q2_diversity/tests/test_beta.py
```diff
@@ -114,24 +115,31 @@ def test_beta_empty_table(self):
...
+                                                 [0.25, 0.00, 0.00],
+                                                 [0.25, 0.00, 0.00]],
+                                                ids=['S1', 'S2', 'S3'])
+
+                actual = actual[0].view(skbio.DistanceMatrix)
+                self.assertEqual(actual.ids, expected.ids)
+                for id1 in actual.ids:
+                    for id2 in actual.ids:
+                        npt.assert_almost_equal(actual[id1, id2],
+                                                expected[id1, id2])
```
ebolyen, line 142: Could we refactor this into a test helper?

q2_diversity/tests/test_beta.py
```diff
@@ -114,24 +115,31 @@ def test_beta_empty_table(self):
...
-        actual = actual[0].view(skbio.DistanceMatrix)
-        self.assertEqual(actual.ids, expected.ids)
-        for id1 in actual.ids:
-            for id2 in actual.ids:
-                npt.assert_almost_equal(actual[id1, id2], expected[id1, id2])
+                # expected computed with skbio.diversity.beta_diversity
+                expected = skbio.DistanceMatrix([[0.00, 0.25, 0.25],
+                                                 [0.25, 0.00, 0.00],
+                                                 [0.25, 0.00, 0.00]],
+                                                ids=['S1', 'S2', 'S3'])
```
ebolyen, line 135: This can probably sit outside the subtest loop, since it never changes.

q2_diversity/tests/test_beta.py
```diff
@@ -153,230 +166,332 @@ def test_beta_phylogenetic_unknown_metric(self):
...
+                     0.78397813, 0.72318399, 0.76138933, 0.61041275,
+                     0.62331299, 0.71848305, 0.70416337, 0.75258475,
+                     0.79249029, 0.64392779, 0.70052733, 0.69832716,
+                     0.77818938, 0.72959894, 0.75782689, 0.71005144,
+                     0.75065046, 0.78944369, 0.63593642, 0.71283615,
+                     0.58314638, 0.69200762, 0.68972056, 0.71514083])
+                ids = ('10084.PC.481', '10084.PC.593', '10084.PC.356',
+                       '10084.PC.355', '10084.PC.354', '10084.PC.636',
+                       '10084.PC.635', '10084.PC.607', '10084.PC.634')
+                expected = skbio.DistanceMatrix(data, ids=ids)
```
ebolyen, line 218: Same here, can be moved outside the loop

q2_diversity/tests/test_beta.py
```diff
@@ -153,230 +166,332 @@ def test_beta_phylogenetic_unknown_metric(self):
...
+
+                # id order can be different with _meta, so order w.r.t.
+                # expected.
+                actual = actual.filter(ids)
+
+                self.assertEqual(actual.ids, expected.ids)
+                for id1 in actual.ids:
+                    for id2 in actual.ids:
+                        npt.assert_almost_equal(actual[id1, id2],
+                                                expected[id1, id2])
```
ebolyen, line 233: ditto: test helper

ebolyen (changes requested):
Looks good to me, but I think the tests can be cleaned up pretty easily, just some minor changes there.

ebolyen:
I would also second @ChrisKeefe's point on plural for the inputs

wasade:
For the inputs, the user is entering `--i-table <path> --i-table <path> ...` as I don't think there is a good way to describe multiple entries for a single argument right now. I think `table` makes sense here as other actions which take, for example, multiple metadata files I don't believe rely on the plural form

q2_diversity/plugin_setup.py
ebolyen: Turns out I needed coffee, I think the only constraint here is whether P/A can be used in combination with the others, which means this would probably be fine:

```ipytb
In [10]: T_table, P_metric, _ = TypeMap({
    ...:     (List[FeatureTable[Frequency | RelativeFrequency | PresenceAbsence]],
    ...:      Str % Choices("unweighted_unifrac")): Visualization,  # a convention for no-op w.r.t. output
    ...:     (List[FeatureTable[Frequency | RelativeFrequency]],
    ...:      Str % Choices('weighted_unifrac', 'weighted_normalized_unifrac', 'generalized_unifrac')): Visualization
    ...: })

In [11]: T_table
Out[11]: List[FeatureTable[Frequency | RelativeFrequency | PresenceAbsence]]¹ | List[FeatureTable[Frequency | RelativeFrequency]]²

In [12]: P_metric
Out[12]: Str % Choices('unweighted_unifrac')¹ | Str % Choices('weighted_unifrac', 'weighted_normalized_unifrac', 'generalized_unifrac')²
```

It is less hideous...

thermokarst:
I agree with @ChrisKeefe & @ebolyen, I think `tables` and `phylogenies` would be more clear (and if we update, we should fix that over in q2-div-lib before the release ships).

`--i-tables table1.qza table2.qza table3.qza` and `--i-phylogenies tree1.qza tree2.qza tree3.qza` looks pretty nice to me.

q2_diversity/plugin_setup.py
wasade: Do you mean whether to allow a user to specify `--i-table a_frequency_table.qza --i-table a_presenceabsence_table.qza ...`? I don't believe there is enough data to guide whether this should or should not be restricted like this

ebolyen:
You can actually do this now:

```
--i-tables <path1> <path2> <path3>
```

(I got tired of not being able to do `data/*` as input...)

Other list-likes use plural as well.

ebolyen:
@thermokarst jinx!

q2_diversity/tests/test_beta.py
wasade: Sure. Note that most of this is not new, just indented, but it could definitely benefit from some refactor. Will clean up

wasade:
Oh, nice!!! I think that's new from when we discussed a few months back? I agree much cleaner, and agree with using the plural form

thermokarst:
> if users are likely to pass Lists of them

I think that that is the point of the method, right @wasade? If that's the case, do we need to validate the length? Does it make sense to run only on a single table/tree pair?

thermokarst:
> I think that's new from when we discussed a few months back?

@wasade, this functionality has been around since mid-2019. I think what we were discussing a few months back was a syntax for making these pairs full-fledged tuples, explicitly saying "this table belongs with this tree."

wasade:
@thermokarst, ah okay. I thought what I put in here was what was the recommendation from that chat :) but I must have misread. The "most correct" thing would be to validate that each tree/table in a common index position correspond to each other or if QIIME2 was able to ensure the paired relationships. Operating on a single tree/table is fine and is identifical to using regular unifrac

thermokarst:
Discussed out of band with @wasade & @ebolyen, going to postpone this q2-diversity integration until a future release. In the meantime, the new `unifac.meta` functionality can be found in q2-diversity-lib (and will be published in the 2020.11 core distribution).

Some remaining questions to solve: do we expose this `meta` functionality in its own pipeline, do we expose it via the existing `beta-phylogenetic` pipeline, or do we only leave it exposed through q2-diversity-lib?

q2_diversity/plugin_setup.py
thermokarst: > Quick question, can the list be polymorphic as described, or should it be monomorphic like List[FeatureTable[Frequency] | FeatureTable[RelativeFrequency] | FeatureTable[PresenceAbsence]] ?

A quick note on this - only one of these are actually possible here, because the diversity-lib `beta-phylogenetic-meta-passthrough` only works with `FeatureTable[Frequency]`.

wasade: Okay, though we may want to consider relaxing the underlying library

thermokarst: Filed an issue here: https://github.com/qiime2/q2-diversity-lib/issues/30

lizgehret:
Hey @ebolyen circling back here - should we revisit this?

ebolyen:
This one's definitely quite old. @wasade do you or your team have any preferences on the matter?

Rereading the thread, it seems like this is all already implemented in q2-div-lib, so this PR is mostly just copying it into the "diversity" namespace so to speak.

I'm inclined to close it, since it hasn't really come up since and it may make more sense to modify `beta-phylogenetic` at some point instead. Please feel free to re-open at any point :)

wasade:
It would be very exciting for `meta` to be available to the user community. Is it now possible for people to use it? If not, then this should be reopened and we should work toward merge.

ebolyen:
Yep! It's already available in the default install as [`qiime diversity-lib beta-phylogenetic-meta-passthrough`](https://docs.qiime2.org/2022.8/plugins/available/diversity-lib/beta-phylogenetic-meta-passthrough/). It's perhaps not ideally named, but it's definitely been around for a while now.

If we'd like it in this particular diversity plugin it may make more sense to put it in `beta-phylogenetic` outright, since that's probably what a user would expect. Although the inputs might get tricky to explain.

wasade:
Is that easy to do?

ebolyen:
My hunch is no, because we would need to account for the different input type signatures here, list vs singular.

I know we can't take a union of `List[X] | X` so either that changes, or typemap accidentally does let this work. But then we'd have to double check that the interfaces can tolerate that.

Alternatively, we make the breaking change for `beta-phylogenetic` to the input parameters and update various call-sites like `core-metrics-phylogenetic`.

---

What if we rename it in diversity-lib so that it doesn't have "passthrough", which isn't particularly illuminating for end-users? That would be fast and a breaking change that is unlikely to rile anyone up.

wasade:
I'm in favor of easy :) We have students actively exploring this method right now directly via the unifrac API, and its historical precedence is impressive, so simplifying use is I think a bit + for the user base

ebolyen:
Would one of them be able to create a PR to update the name (dropping 'passthrough') and fix this issue: https://github.com/qiime2/q2-diversity-lib/issues/30?

This could make a good first-PR for someone.

wasade:
@ahdilmore, would you be interested in sorting this out? Basically, this would be to help make UniFrac's meta method more readily accessible via QIIME 2

PR #300, opened 2020-11-18
Added t-SNE and UMAP Function

TimJLU:
I added t-SNE Function to the _ordination.py File.

It is just a first draft to start a Discussion and still in progress.

Next step would be:
Better way to pass Arguments.
Create a default setting if you don't pass any Arguments.


sjanssen2:
This PR is a result of our discussion in the forum: https://forum.qiime2.org/t/t-sne-or-umap-as-q2-plugins/14477

sjanssen2:
I wonder why Travis isn't trigged by this PR - are we missing something here? Or do we have to be more patient?

thermokarst:
> I wonder why Travis isn't trigged by this PR - are we missing something here?

I have no clue why it didn't trigger... Maybe GitHub had a hiccup? EIther way, I don't expect the tests to tell us to much, since the functional here hasn't been wired into the plugin, and there are no unit tests yet.

This is pretty exciting, let us know what you need from us!

q2_diversity/_ordination.py
```diff
@@ -28,3 +30,48 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+    data = distance_matrix.data
+    ids = distance_matrix.ids
+    umap_data = reducer.fit_transform(data)
+    axis_labels = ["UMAP%d" % i for i in range(1, n_components + 1)]
+    eigenvalues = [0 for i in axis_labels]
+    long_method_name = "Uniform Manifold Approximation and Projection"
+
+    return skbio.OrdinationResults(
+        short_method_name="UMAP",
+        long_method_name=long_method_name,
```
sjanssen2, line 71: why is it necessary to use a variable in line 67 for the name?

TimJLU: It's not I just tried to stay uniform with the existing code. I can chane that.

.travis.yml
```diff
@@ -17,6 +17,7 @@ install:
   - pip install -q flake8 coveralls
   - pip install -q https://github.com/qiime2/q2lint/archive/master.zip
   - make install
+  - conda install -c conda-forge umap-learn
```
TimJLU, line 20: is that the best way to add the conda package to the installation.

q2_diversity/__init__.py
```diff
@@ -23,8 +23,8 @@
 
 
 __all__ = ['beta', 'beta_phylogenetic', 'alpha', 'alpha_phylogenetic',
-           'pcoa', 'pcoa_biplot', 'alpha_group_significance', 'bioenv',
-           'beta_group_significance', 'alpha_correlation',
+           'pcoa', 'tsne', 'uMAP' 'pcoa_biplot', 'alpha_group_significance',
```
sjanssen2, line 26: looks like a `,` is missing here

TimJLU: oh yeah! Thanks for the quick reply

.travis.yml
thermokarst: The `umap-learn` dependency should be added to the package recipe, https://github.com/qiime2/q2-diversity/blob/cb84424cd3707432e63ed400a1b90345dc8d2e22/ci/recipe/meta.yaml#L21, and removed from this `.travis.yml` config.

q2_diversity/_ordination.py
```diff
@@ -5,26 +5,111 @@
...
 import skbio.stats.ordination
 import pandas as pd
+import umap
 
 
-def pcoa(distance_matrix: skbio.DistanceMatrix,
-         number_of_dimensions: int = None) -> skbio.OrdinationResults:
+def pcoa(
+    distance_matrix: skbio.DistanceMatrix, number_of_dimensions: int = None
+) -> skbio.OrdinationResults:
```
thermokarst, line 16: There are a bunch of stylistic changes like this in this PR. I don't think they're necessary, and they add a bit of noise to the PR diff - can you please undo any of these changes that aren't actually implementing any new features, etc? Thanks!

.travis.yml
TimJLU: okay, thank you for the hint!

q2_diversity/_ordination.py
thermokarst: Thanks for addressing! To be clear, I wasn't requesting that you change anything about the new code contributed here ("added the old code style to new tsne and umap function"), I just wanted the broad-reaching stylistic changes reverted. It pollutes the diff, git history, and the git blame. You're welcome to introduce whatever style makes sense to you with the new code, we just ask that you not start changing _existing_ code, without a clear reason. Thanks!

TimJLU: Yeah, the changes came from Flake8 which I installed on my code editor (Visual Studio Code). I had no intentions to change something from the old code."Old Style" was just a poor way to say "previous Style". I'm new to Git Hub and not so familiar with the correct Terms like WIP and how to adress commits probably so I'm very glad for every Feedback that I get

TimJLU:
The recent check has failed, but I don´t get the Error Message. Is it because of the tsne Function and the way I used **kwargs?

q2_diversity/_ordination.py
```diff
@@ -28,3 +30,80 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
 def pcoa_biplot(pcoa: skbio.OrdinationResults,
                 features: pd.DataFrame) -> skbio.OrdinationResults:
     return skbio.stats.ordination.pcoa_biplot(pcoa, features)
+
+
+def tsne(distance_matrix: skbio.DistanceMatrix,
+         **kwargs) -> skbio.OrdinationResults:
```
thermokarst, line 36:
```suggestion
def tsne(distance_matrix: skbio.DistanceMatrix) -> skbio.OrdinationResults:
```

`**kwargs` is invalid when defining a QIIME 2 Action - you need specify a parameter for every argument. This is so that the QIIME 2 framework can build the interface that matches the signature.

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,77 @@
...
+                                "of which varies across different datasets."
+    },
+    output_descriptions={'tsne': 'The resulting t-SNE matrix.'},
+    name='t-distributed stochastic neighbor embedding',
+    description=("Apply t-distributed stochastic neighbor embedding."),
+
+)
+
+plugin.methods.register_function(
+    function=q2_diversity.uMAP,
```
sjanssen2, line 280: I think it would be more consistent to use all lowercase character for the function name, i.e. umap instead of uMAP. Same applies for the parameter regarding the output artifact

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,55 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+        tsneData = TSNE().fit_transform(data)
+
+    axis_labels = ["TSNE%d" % i for i in range(1, number_of_dimensions + 1)]
+    eigenvalues = [0 for i in axis_labels]
+
+    return skbio.OrdinationResults(
+        short_method_name="T-SNE",
+        long_method_name="t-distributed stochastic neighbor embedding",
+        eigvals=pd.Series(eigenvalues, index=axis_labels),
+        samples=pd.DataFrame(tsneData, index=ids, columns=axis_labels),
```
sjanssen2, line 58: Emperor expect a non empty proportion_explained attribute in OrdinationResults. Is this computed via tsne? If not, `proportion_explained=pd.Series(None, index=axis_labels)` might do the trick here

q2_diversity/plugin_setup.py
sjanssen2: but beware of name clashed between sklearn umap an qiime umap

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,55 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+
+    axis_labels = ["UMAP%d" % i for i in range(1, number_of_dimensions + 1)]
+    eigenvalues = [0 for i in axis_labels]
+
+    return skbio.OrdinationResults(
+        short_method_name="UMAP",
+        long_method_name="Uniform Manifold Approximation and Projection",
+        eigvals=pd.Series(eigenvalues, index=axis_labels),
+        samples=pd.DataFrame(umap_data, index=ids, columns=axis_labels),
+    )
```
sjanssen2, line 85: proportion_explained also missing

q2_diversity/plugin_setup.py
TimJLU: "but beware of name clashed between sklearn umap an qiime umap" 

Thats why I used mixed cases but I'll rename the import to up so umap is available for the function

q2_diversity/_ordination.py
TimJLU: Thank you very much :)

sjanssen2:
We should think about which parameters of the algorithms should be exposed to the user via the plugin. I'd think exposing **all** parameters is overkill and will confuse the default user. Unfortunately, I am not very familiar with both techniques and cannot answer this question due to personal experience :-/

### t-SNE
Reading about t-SNE, I think our plugin should expose `perplexity` and `n_components`. If the Qiime2 framework allows access to the `--verbose` parameter, we might also want to pass that to the `verbose` flag of the sklearn algorithm. And since we will always pass a precomputed distance matrix, `n_jobs` should be set to 1. Not sure if we want to expose it as `--p-n-jobs` to indicate that multiple CPUs/cores won't speed up the process?

### UMAP
UMAP documentation has the great chapter https://umap-learn.readthedocs.io/en/latest/parameters.html# which tells me we should expose `n_neighbors`, `min_dist` and of course `n_components`

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,57 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+    ids = distance_matrix.ids
+
+    if number_of_dimensions == 2:
+        number_of_dimensions = 3
+        tsne = TSNE(2).fit_transform(data)
+        add_zeros = np.zeros((tsne.shape[0], 1), dtype=np.int64)
+        tsneData = np.append(tsne, add_zeros, axis=1)
+
+    else:
+        tsneData = TSNE(number_of_dimensions).fit_transform(data)
```
sjanssen2, line 49: can you use a named parameter here for clarity, i.e. `TSNE(n_components=number_of_dimension)`?

q2_diversity/_ordination.py
```diff
@@ -34,19 +34,25 @@ def pcoa_biplot(pcoa: skbio.OrdinationResults,
 
 
 def tsne(distance_matrix: skbio.DistanceMatrix,
-         number_of_dimensions: int = 3) -> skbio.OrdinationResults:
+         number_of_dimensions: int = 3,
```
sjanssen2, line 37: I think we should set the default value to 2, since this is the more common use of t-SNE - although PCoA reduces to three dimensions by default

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,67 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+        short_method_name="T-SNE",
+        long_method_name="t-distributed stochastic neighbor embedding",
+        eigvals=pd.Series(eigenvalues, index=axis_labels),
+        proportion_explained=pd.Series(None, index=axis_labels),
+        samples=pd.DataFrame(tsneData, index=ids, columns=axis_labels),
+    )
+
+
+def umap(distance_matrix: skbio.DistanceMatrix,
+         number_of_dimensions: int = 3,
```
sjanssen2, line 70: same as above: default should be 2

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,67 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+    data = distance_matrix.data
+    ids = distance_matrix.ids
+
+    if number_of_dimensions == 2:
+        number_of_dimensions = 3
+        reducer = up.UMAP(n_components=2).fit_transform(data)
+        add_zeros = np.zeros((reducer.shape[0], 1), dtype=np.int64)
+        umap_data = np.append(reducer, add_zeros, axis=1)
+    else:
+        reducer = up.UMAP(n_components=number_of_dimensions,
```
sjanssen2, line 83: reads like a code duplication. Should we wrap this filling of the third dimension into a private function and call from here and within tsne?

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,85 @@
...
+                                "and eigenvalues in an exact manner. For very "
+                                "large matrices, this is expected to be slow. "
+                                "If a value is specified for this parameter, "
+                                "then the fast, heuristic "
+                                "eigendecomposition algorithm fsvd "
+                                "is used, which only computes and returns the "
+                                "number of dimensions specified, but suffers "
+                                "some degree of accuracy loss, the magnitude "
+                                "of which varies across different datasets.",
+        'perplexity':           ("Text will follow"),
```
sjanssen2, line 274: ähm!

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
     citations=[citations['legendrelegendre']]
 )
 
+plugin.methods.register_function(
+    function=q2_diversity.tsne,
+    inputs={'distance_matrix': DistanceMatrix},
+    parameters={
+        'number_of_dimensions': Int % Range(1, None),
```
thermokarst, line 247: Is 1 a valid value for this method? The implementation code doesn't handle the case where 1 is specified, and I imagine anything less than 2 doesn't make much sense. If that is correct, please update this parameter:

```suggestion
        'number_of_dimensions': Int % Range(2, None),
```

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+                            'computed.')
+    },
+    parameter_descriptions={
+        'number_of_dimensions': "Dimensions to reduce the distance matrix to.",
+        'perplexity':           "Provide the Balance between local and global"
+                                "structure. Low values concentrate on local"
+                                "structure. Large values sacrafice local "
+                                "details for a broader global embedding."
+                                "The default value is 25 to achive  better"
+                                "results for small microbiome datasets",
```
thermokarst, line 265:
```suggestion
        'perplexity':           "Provide the balance between local and global"
                                "structure. Low values concentrate on local "
                                "structure. Large values sacrifice local "
                                "details for a broader global embedding. "
                                "The default value is 25 to achieve better "
                                "results for small microbiome datasets.",
```

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+        'early_exaggeration':   "Affects the tightnes of the shown clusters"
+                                "Larger values increase the distance between"
+                                "natural clusters in the embedded space",
```
thermokarst, line 268:
```suggestion
        'early_exaggeration':   "Affects the tightness of the shown clusters. "
                                "Larger values increase the distance between "
                                "natural clusters in the embedded space.",
```

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+        'learning_rate':        "controls how much the weights are adjusted "
+                                "at each update",
```
thermokarst, line 270:
```suggestion
        'learning_rate':        "Controls how much the weights are adjusted "
                                "at each update.",
```

q2_diversity/_ordination.py
```diff
@@ -28,3 +31,76 @@ def pcoa(distance_matrix: skbio.DistanceMatrix,
...
+                          n_neighbors=n_neighbors,
+                          min_dist=min_dist)
+        reducer = reducer.fit_transform(data)
+        add_zeros = np.zeros((reducer.shape[0], 1), dtype=np.int64)
+        umap_data = np.append(reducer, add_zeros, axis=1)
+    else:
+        reducer = up.UMAP(n_components=number_of_dimensions,
+                          n_neighbors=n_neighbors,
+                          min_dist=min_dist)
+        umap_data = reducer.fit_transform(data)
```
thermokarst, line 95: Same comment as above - it looks like you can compute UMAP outside of a branch, and instead just handle the padding in its own branch.

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+    name='t-distributed stochastic neighbor embedding',
+    description=("Apply t-distributed stochastic neighbor embedding."),
+
+)
+
+plugin.methods.register_function(
+    function=q2_diversity.umap,
+    inputs={'distance_matrix': DistanceMatrix},
+    parameters={
+        'number_of_dimensions': Int % Range(1, None),
```
thermokarst, line 282: Same question as above - should the minimum allowed value be 2?

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+    input_descriptions={
+        'distance_matrix': ('The distance matrix on which UMAP should be '
+                            'computed.')
+    },
+    parameter_descriptions={
+        'number_of_dimensions': "Dimensions to reduce the distance matrix to.",
+        'n_neighbors':          "Provide the Balance between local and global"
+                                " structure. Low Values concentrate on local"
+                                "structure. Large Values sacrafice local "
+                                "details for a broader global embedding",
```
thermokarst, line 296: Is "concentrate" the right word here?

```suggestion
        'n_neighbors':          "Provide the balance between local and global"
                                " structure. Low values concentrate on local "
                                "structure. Large values sacrifice local "
                                "details for a broader global embedding",
```

q2_diversity/plugin_setup.py
```diff
@@ -240,6 +240,71 @@
...
+        'min_dist':             "controls the Cluster Size. Low Values cause "
+                                "clumpier CLusters. Higher Values preserve a"
+                                "broad topological structure. To get"
+                                "less overlapping data points the min_dist"
+                                "default value is set to 0,4",
```
thermokarst, line 301: We might need a reference for "clumpier clusters" somewhere in here - I'm not sure what it means.

```suggestion
        'min_dist':             "Controls the cluster size. Low values cause "
                                "clumpier clusters. Higher values preserve a "
                                "broad topological structure. To get "
                                "less overlapping data points the "
                                "default value is set to 0.4.",
```

q2_diversity/tests/test_ordination.py
```diff
@@ -52,3 +53,37 @@ def test_pcoa_biplot(self):
...
+                                        [0.3333333, 0.0000000, 0.4285714],
+                                        [0.6666667, 0.4285714, 0.0000000]],
+                                       ids=['S1', 'S2', 'S3'])
+
+    def test_tsne_default(self):
+        observed = tsne(self.dm).samples.shape
+        self.assertEqual(observed, (3, 3))
+
+    def test_tsne__custom(self):
+        observed_second = tsne(self.dm, 2).samples.shape
```
thermokarst, line 71: Is this test testing the exact same path as the test above? I'm a little confused, because the default number of dimensions is `2` (as specified in your implementation function, above). Perhaps this test should be testing the number of dimensions value of `3`, instead?

q2_diversity/tests/test_ordination.py
```diff
@@ -52,3 +53,37 @@ def test_pcoa_biplot(self):
...
+        self.dm = skbio.DistanceMatrix([[0.0000000, 0.3333333, 0.6666667],
+                                        [0.3333333, 0.0000000, 0.4285714],
+                                        [0.6666667, 0.4285714, 0.0000000]],
+                                       ids=['S1', 'S2', 'S3'])
+
+    def test_tsne_default(self):
+        observed = tsne(self.dm).samples.shape
+        self.assertEqual(observed, (3, 3))
+
+    def test_tsne__custom(self):
```
thermokarst, line 70:
```suggestion
    def test_tsne_custom_number_of_dims(self):
```

q2_diversity/tests/test_ordination.py
```diff
@@ -52,3 +53,37 @@ def test_pcoa_biplot(self):
...
+
+    def setUp(self):
+        self.dm = skbio.DistanceMatrix([[0.0000000, 0.3333333, 0.6666667],
+                                        [0.3333333, 0.0000000, 0.4285714],
+                                        [0.6666667, 0.4285714, 0.0000000]],
+                                       ids=['S1', 'S2', 'S3'])
+
+    def test_tsne_default(self):
+        observed = tsne(self.dm).samples.shape
+        self.assertEqual(observed, (3, 3))
```
thermokarst, line 68: Is there anything else we can test here, besides the dimensions? Is this test stochastic? If not, perhaps you can add an assertion that verifies the right values were calculated.

q2_diversity/tests/test_ordination.py
```diff
@@ -52,3 +53,37 @@ def test_pcoa_biplot(self):
 
         skbio.util.assert_ordination_results_equal(observed, expected,
                                                    ignore_directionality=True)
+
+
+class tsneTests(unittest.TestCase):
```
thermokarst, line 58: A general note - we tend to use the "arrange - act - assert" pattern with respect to unit tests. You don't need to rewrite anything here, but I wanted to show you what this might look like - we like it, and think its a useful pattern:

```def test_tsne_default(self):
    observed = tsne(self.dm)

    self.assertEqual(observed.samples.shape, (3, 3))
    self.assertEqual(something else...)
```

q2_diversity/tests/test_ordination.py
```diff
@@ -52,3 +53,37 @@ def test_pcoa_biplot(self):
...
+        np.fill_diagonal(data, 0)
+        self.dm = skbio.DistanceMatrix(data,
+                                       ids=['S1', 'S2', 'S3', 'S4', 'S5'])
+
+    def test_umap_default(self):
+        observed = umap(self.dm).samples.shape
+        self.assertEqual(observed, (5, 3))
+
+    def test_umap_custom(self):
+        observed_second = umap(self.dm, 2).samples.shape
```
thermokarst, line 88: Same question as above - is this testing the same thing as the test above?

thermokarst (changes requested):
Thanks for the submission, @TimJLU! I have suggested a handful of changes, as well as some copy-edits, and other misc questions (inline). I also have a few broader questions and comments:

- Please don't forget to add any relevant citations into your new method definitions.
- Right now the `number_of_dimensions` parameters are _required_ (but since they specify a default value of 2, you don't necessary have to provide that as input). Looking at the `pcoa` method in this plugin, it looks like there we have made the `number_of_dimensions` parameter completely _optional_. I am curious why you aren't doing that here - care to explain?
- Do we need a guard in here that ensures that the number of dimensions is greater than or equal to the length of the input distance matrix?

Thanks so much, this'll be really exciting to get in!

PS - ccing @sjanssen2 - I am not necessarily the right person to review the scientific soundness of these methods, I'll leave that level of the review up to you.

sjanssen2:
many thanks @thermokarst. @TimJLU is on a writing sprint to finish his masters thesis. It contains some benchmark and examples for applying PCoA, tSNE, umap to different datasets. Hope we can distil his results into a nice little summary to show that each method suits different purposes and therefore it is great to have them all in qiime2

TimJLU:
I finally were able to implement all your requested changes @thermokarst. 

> Right now the number_of_dimensions parameters are required (but since they specify a default value of 2, you don't necessary have to provide that as input). Looking at the pcoa method in this plugin, it looks like there we have made the number_of_dimensions parameter completely optional. I am curious why you aren't doing that here - care to explain?

During my thesis I got the best visual results for t-SNE and UMAP by reducing the dimensions to 2. The idea was to use recommended default settings so that inexperienced Users could get the best results without changing any of the variables.

q2_diversity/_version.py
```diff
@@ -23,9 +23,9 @@ def get_keywords():
     # setup.py/versioneer.py will grep for the variable names, so they must
     # each be defined on a line of their own. _version.py will just call
     # get_keywords().
-    git_refnames = "$Format:%d$"
-    git_full = "$Format:%H$"
-    git_date = "$Format:%ci$"
+    git_refnames = " (tsneupload)"
+    git_full = "7d64dcdf1efff39a893f3a3ab132d159bcc25b19"
+    git_date = "2021-05-03 11:12:26 +0200"
```
ebolyen, line 28: This shouldn't have changed as it's an internal detail to packaging for the plugin.

ci/recipe/meta.yaml
```diff
@@ -46,6 +46,7 @@ requirements:
     - q2-emperor {{ release }}.*
     - q2-diversity-lib {{ release }}.*
     - r-vegan >=2.5_3
+    - umap-learn
```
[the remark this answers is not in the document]
ebolyen: I don't suppose the situation has changed in the last few months?

ebolyen (changes requested):
Hi @TimJLU, hope your thesis went well! Could you resolve some of the outstanding questions @thermokarst had (there's a few still left)? Overall super excited about this functionality!

thermokarst (commented):
Thanks for the review @ebolyen - I apparently had a few review comments queued up, I have no clue how old these are, but I'll send 'em out into the world.

ci/recipe/meta.yaml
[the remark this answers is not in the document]
TimJLU: I had no issues with the default numba version. So at the moment I'm not quite sure how to deal with that problem. How can I change the version of single package within the umap-learn package and apply it to the meta.yaml? Any suggestions?

q2_diversity/plugin_setup.py
TimJLU: Yes. You are absolutely right! Theoretically, it is possible, but I don't see any use cases where it would be beneficial to present or explore data in just one dimension.

q2_diversity/tests/test_ordination.py
TimJLU: Different runs with the same matrix can produce different results(for both t-SNE and UMAP). Variance between runs are relatively small but it is not possible to predict the exact results.

TimJLU:
Thank you very much @ebolyen and @thermokarst. I hope I changed everything accordingly. The only question still up is how to deal with the numba package within in the umap-learn package? Did I miss something?

TimJLU:
Hey, i saw that the numba package got an update on August 20. Maybe this would solve the problem on it's own?

gregcaporaso:
@TimJLU, thanks so much for this pull request and apologies for how long it took us to get this in! 

I just pulled this branch down and tested both new actions with the Moving Pictures tutorial data and a new tutorial data set that we're working with ([Liao et al, 2021](https://www.nature.com/articles/s41597-021-00860-8)). The umap-learn team's conda install command worked for me, so it looks like that issue is sorted out, and the plots I generated all look reasonable. We're going to run the tests, and as long as everything seems to be working we'll get this merged for the 2021.11 release (scheduled for next week).

ebolyen (approved):

thermokarst (approved):

PR #303, opened 2021-02-15
imp: adonis - tighten scope of nan-checking; tests

thermokarst:
This is a follow-up PR for https://github.com/qiime2/q2-diversity/pull/282

cc @collin-le & @nbokulich

---

This PR tightens up the NaN checking to _only the columns used in the formula_, rather than raising an error on _any_ column, even if it isn't used.

This changeset also adds a few quick unit tests to exercise this behavior.

### DIFF
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -363,11 +363,11 @@ def adonis(output_dir: str,
     terms = ModelDesc.from_formula(formula)
     for t in terms.rhs_termlist:
         for i in t.factors:
-            metadata.get_column(i.name())
-
-    metadata_df = metadata.to_dataframe()
-    if metadata_df.isnull().values.any():
-        raise ValueError("Adonis cannot run with NaNs in metadata.")
+            column = metadata.get_column(i.name())
+            if column.has_missing_values():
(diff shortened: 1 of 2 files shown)


nbokulich (approved):
All looks good to me, thanks @thermokarst ! Nothing to add here, just an open question: should we (here or in a future PR) add an option to drop NaNs instead of failing? We have similar options elsewhere in q2-div and beyond and could be convenient.

Some tests are failing locally but this does not appear related to this PR.

thermokarst:
> add an option to drop NaNs instead of failing?

I'm game - but how about we do it in a future dev cycle? I just wanted to wrap up @collin-le's work on this in time for the upcoming release next week.

> Some tests are failing locally but this does not appear related to this PR.

Can you share some logs? I'm all green here.

nbokulich:
> but how about we do it in a future dev cycle?

👍 opening an issue momentarily...

> Can you share some logs? I'm all green here.

I can share out of loop but the error is clearly unrelated, coming from `core-metrics`. I am going to go ahead and hit MERGE 🟢

thermokarst:
> I can share out of loop but the error is clearly unrelated, coming from core-metrics

There was a huge shuffle this summer related to q2-div-lib - you might need to refresh your dev env, in case you haven't in a while.

PR #306, opened 2021-03-30
imp: m2 fmt validation and metadata-compatible

thermokarst:
@ahdilmore - this PR tweaks the format of the M2 stats a little bit, to make it compatible with QIIME 2 Metadata - this will allow users to "view" the output stats using q2-metadata's `tabulate` visualizer. Basically I transposed the Pandas table, and made the file format into a TSV. I added a new transformer to facilitate viewing as Metadata, and also added some code for validating the format. Would you mind reviewing this and letting me know what you think? Thanks so much! 🦖 

PS - I also cleaned up some parameter description formatting and tacked it into this changeset.

### DIFF
--- a/q2_diversity/_format.py
+++ b/q2_diversity/_format.py
@@ -7,11 +7,28 @@
 # ----------------------------------------------------------------------------
 
 import qiime2.plugin.model as model
+from qiime2.plugin import ValidationError
+import qiime2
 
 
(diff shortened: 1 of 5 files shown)


ahdilmore:
Missed @ElDeveloper's last comment on there - let's go with integrating the changes into q2-emperor first, then circling back to this PR.

PR #308, opened 2021-04-15
ENH: Multiple changes and enhancements

ElDeveloper:
- Adds a permutations parameter.
- Add handling for zero permutations.
- Avoid returning p-value=0 to mimic the behaviour of other
  permutation-based tests.
- Rename trials to permutations.

### DIFF
--- a/q2_diversity/_procrustes.py
+++ b/q2_diversity/_procrustes.py
@@ -86,6 +86,9 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
 
     trials_below_m2 = 0
 
+    if permutations == 'disable':
+        permutations = 0
+
     for i in range(permutations):
 
         # shuffle rows in np array
@@ -98,7 +101,7 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
         if m2 < true_m2:
             trials_below_m2 += 1
 
-    if permutations == 'disable':
+    if permutations == 0:
         p_val = np.nan
     else:
         # mimic the behaviour in scikit-bio's permutation-based tests and avoid
--- a/q2_diversity/_procrustes.py
+++ b/q2_diversity/_procrustes.py
@@ -98,7 +98,7 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
         if m2 < true_m2:
             trials_below_m2 += 1
 
-    if permutations == 0:
+    if permutations == 'disable':
         p_val = np.nan
     else:
         # mimic the behaviour in scikit-bio's permutation-based tests and avoid
--- a/q2_diversity/_procrustes.py
+++ b/q2_diversity/_procrustes.py
@@ -15,9 +15,10 @@ from numpy.random import default_rng
 
 
 def procrustes_analysis(reference: OrdinationResults, other: OrdinationResults,
-                        dimensions: int = 5) -> (OrdinationResults,
-                                                 OrdinationResults,
-                                                 pd.DataFrame):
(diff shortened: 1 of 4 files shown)


q2_diversity/plugin_setup.py
```diff
@@ -251,7 +251,10 @@
 plugin.methods.register_function(
     function=q2_diversity.procrustes_analysis,
     inputs={'reference': PCoAResults, 'other': PCoAResults},
-    parameters={'dimensions': Int % Range(1, None)},
+    parameters={
+        'dimensions': Int % Range(1, None),
+        'permutations': Int % Range(0, None)
```
thermokarst, line 256:
```suggestion
        'permutations': Int % Range(1, None) | Str % Choices('disable')
```

q2_diversity/_procrustes.py
```diff
@@ -100,11 +98,16 @@ def _procrustes_monte_carlo(reference: np.ndarray,
         if m2 < true_m2:
             trials_below_m2 += 1
 
-    p_val = trials_below_m2 / trials
+    if permutations == 0:
```
thermokarst, line 101:
```suggestion
    if permutations == 'disable':
```

q2_diversity/plugin_setup.py
```diff
@@ -260,9 +263,18 @@
...
         'reference': ('The ordination matrix to which data is fitted to.'),
         'other': ("The ordination matrix that's fitted to the reference "
-                  "ordination.")
+                  "ordination."),
+    },
+    parameter_descriptions={
+        'dimensions': ('The number of dimensions to use when fitting the two '
+                       'matrices'),
+        'permutations': 'The number of permutations to be run when computing '
+                        'p-values. Supplying a value of zero will disable '
```
thermokarst, line 272:
```suggestion
                        'p-values. Supplying a value of `disable` will disable '
```

q2_diversity/tests/test_procrustes.py
```diff
@@ -94,6 +109,45 @@ def test_procrustes(self):
...
+        skbio.util.assert_ordination_results_equal(other, self.expected_noise)
+
+        # the p value shouldn't be zero even in the presence of noise
+        self.assertAlmostEqual(true_m2, 0.7388121)
+        self.assertNotAlmostEqual(true_p_value, 0.001)
+
+    def test_zero_permutations_nan_pvalue(self):
+        ref, other, m2_results = procrustes_analysis(self.reference,
+                                                     self.other,
+                                                     permutations=0)
```
thermokarst, line 141:
```suggestion
                                                     permutations='disable')
```

thermokarst (commented):
This is great, @ElDeveloper! I think this should go in with this release (since it changes the format schema, better to get it in now, than later, and have to support multiple format versions). I proposed some changes inline, that get rid of the magic `0` value, and instead use the `'disable'` value instead. In general we've been trying to move away from magic numbers like `0` and `-1` for opting out of specific behaviors.

q2_diversity/plugin_setup.py
ElDeveloper: THIS IS SO NEAT!

ElDeveloper:
Thanks so much @thermokarst I didn't know about this.

ElDeveloper:
Thanks @thermokarst, I was just in the process of fixing the broken test 👍

thermokarst (approved):

thermokarst:
Thanks so much, @ElDeveloper! We'll need to update https://github.com/qiime2/q2-emperor/pull/89 with the new column name, I can do it, but if you get there before me, have at it.

PR #309, opened 2021-04-19
MAINT: Updated q2-diversity from q2-types & syntax adjustments

lizgehret:
Updates corresponding with this PR: https://github.com/qiime2/q2-types/pull/259


ElDeveloper:
Of course! The ordination module. Thanks for updating 🎉

PR #312, opened 2021-08-13
maint: new pandas groupby api

thermokarst:
https://pandas.pydata.org/docs/whatsnew/v1.3.0.html#float-result-for-groupby-mean-groupby-median-and-groupby-var

All groupby operations now return floats. Older versions of the `alpha-rarefaction` viz would return a mix of floats and ints, so this seems nicer all around.


thermokarst:
BTW, the reason we would want floats here, in the first place, is because pandas will take the arithmetic mean of the two "median" elements in the case of an even number of records:

```python
df = pd.DataFrame({
    'a': [False, False, True, True],
    'b': [0, 0, 1, 2],
    'c': [0.0, 0.0, 1.0, 2.0],
}, index=['x', 'x', 'x', 'x'])
df.groupby(df.index).median()
```

```
     a    b    c
x  0.5  0.5  0.5
```

ebolyen (approved):

ebolyen:
@thermokarst how should this get tested, since it seems to be related to the global pins?

PR #316, opened 2021-09-04
Closes `plt.Figure()` after saving mantel test plot to free up memory

crusher083:
In the code, beta correlation `mantel()` in `_visualizer.py` opens an instance of `plt.Figure()` and doesn't close it, consuming memory. Added a line to close the instance and free up memory after saving.


ebolyen (approved):
Nice catch, thanks @crusher083!

ebolyen:
Going to let the tests run on this and then I'll merge :)

PR #318, opened 2021-10-27
ENH/API: support coloring Mantel scatters

wasade:
This pull request introduces a means to color the Mantel scatter plots by a categorical variable, which can help emphasize structure in the plot. An example is attached
Screen Shot 2021-10-27 at 4 12 58 PM

### DIFF
--- a/q2_diversity/_beta/_visualizer.py
+++ b/q2_diversity/_beta/_visualizer.py
@@ -268,6 +268,7 @@ def beta_group_significance(output_dir: str,
 def mantel(output_dir: str, dm1: skbio.DistanceMatrix,
            dm2: skbio.DistanceMatrix, method: str = 'spearman',
            permutations: int = 999, intersect_ids: bool = False,
+           metadata: qiime2.CategoricalMetadataColumn = None,
            label1: str = 'Distance Matrix 1',
            label2: str = 'Distance Matrix 2') -> None:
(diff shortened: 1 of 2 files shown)


lizgehret:
Closed and re-opened to trigger CI workflow

lizgehret:
> That's a much prettier output! Also, the error you're getting confused me for a few minutes:
> 
> ```
> TypeError: Parameter 'metadata' received 'Metadata' as an argument, which is incompatible with parameter type: MetadataColumn[Categorical]
> ```
> 
> But it's complaining about a python string `"Metadata"` which is the default label for beta_correlation (`label1`). I think if you update that pipeline (probably with keywords) the tests should pass.

Hey @wasade I took a quick look over this PR and re-ran ci and we're still getting the same test failures that @ebolyen mentioned above. Our PR cutoff for the 2022.8 release was technically yesterday, so unless you have something ready to go here to address these failures before EOD we'll have to hold off on this one until the 2022.11 release.

wasade:
TBH its been so long I don't recall what this really does. I'm on leave for
a few more weeks, it would be okay to close this.

lizgehret:
Sounds good @wasade, I'll go ahead and close it! We can always re-open if/when you have bandwidth and want to fix up these test failures. Enjoy your time off! 🙂

PR #323, opened 2022-02-16
ENH: add random_state param in dim reduction methods

misialq:
This PR introduces a `random_state` parameter to the two dimensionality reduction methods: tSNE and UMAP for better result reproducibility.

### DIFF
--- a/q2_diversity/_ordination.py
+++ b/q2_diversity/_ordination.py
@@ -38,7 +38,8 @@ def tsne(distance_matrix: skbio.DistanceMatrix,
          perplexity: float = 25.0,
          n_iter: int = 1000,
          learning_rate: float = 200.0,
-         early_exaggeration: float = 12.0) -> skbio.OrdinationResults:
+         early_exaggeration: float = 12.0,
+         random_state: int = 1) -> skbio.OrdinationResults:
 
     data = distance_matrix.data
     ids = distance_matrix.ids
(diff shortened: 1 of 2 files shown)


q2_diversity/_ordination.py
```diff
@@ -38,15 +38,17 @@ def tsne(distance_matrix: skbio.DistanceMatrix,
          perplexity: float = 25.0,
          n_iter: int = 1000,
          learning_rate: float = 200.0,
-         early_exaggeration: float = 12.0) -> skbio.OrdinationResults:
+         early_exaggeration: float = 12.0,
+         random_state: int = 1) -> skbio.OrdinationResults:
```
ebolyen, line 42: This should default to `None` so that it's optional (it's also a valid input for sklearn with the expected semantics).

q2_diversity/_ordination.py
```diff
@@ -68,14 +70,16 @@ def tsne(distance_matrix: skbio.DistanceMatrix,
 def umap(distance_matrix: skbio.DistanceMatrix,
          number_of_dimensions: int = 2,
          n_neighbors: int = 15,
-         min_dist: float = 0.4) -> skbio.OrdinationResults:
+         min_dist: float = 0.4,
+         random_state: int = 1) -> skbio.OrdinationResults:
```
ebolyen, line 74: Same here (umap package also supports `None`).

ebolyen (changes requested):

misialq:
Thanks for pointing those out @ebolyen, all's updated!

ebolyen (approved):
Thanks @misialq!

PR #324, opened 2022-02-22
Citation updates/DOI adds

ChrisKeefe:
Our Halko citation is from [arxiv](https://arxiv.org/abs/1007.5510). Looks like it was published in [SIAM](https://epubs.siam.org/doi/epdf/10.1137/100804139) the following year. Is general best practice to move to the peer-reviewed pub when we notice this?

I've been thinking about DOIs a bit lately, so added DOI fields for a few commonly-used resources. I haven't updated the citation keys for those or anything - that's OK, right?

### DIFF
--- a/q2_diversity/citations.bib
+++ b/q2_diversity/citations.bib
@@ -141,7 +141,8 @@
   number={260},
   pages={583--621},
   year={1952},
-  publisher={Taylor \& Francis}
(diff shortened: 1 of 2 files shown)


ChrisKeefe:
CI seems to be failing during the solving/install process. Is this something we should be concerned about?

thermokarst (approved):

PR #325, opened 2022-03-15
ENH:Add subsample w/replacement in phylo-cm

ElDeveloper:
This is useful for metagenomic studies where there's frequently over 1M
counts per sample. As a side note, the non-phylogenetic variant already has this
parameter.


ebolyen (approved):
Thanks @ElDeveloper!

ElDeveloper:
Thanks!

PR #329, opened 2022-06-03
maint: fix tests in prep from skbio 0.5.7

thermokarst:


thermokarst:
cc @ebolyen - these are the changes we discussed a couple weeks back.

PR #332, opened 2022-10-24
ex: adds usage examples for Hacktoberfest

lizgehret:
This PR is part of our 2022 Hacktoberfest team-wide sprint to add usage examples to all methods in each of our core distribution plugins. The methods examples have been added for are as follows:
```
alpha-group-significance
alpha-correlation
```


Oddant1:
@lizgehret it looks like this PR also needs the URLs updated

Oddant1 (approved):
LGTM

PR #333, opened 2022-11-07
imp: adding R traceback for better error handling

lizgehret:
This PR adds a preliminary code block that will provide an R traceback for errors encountered within this R wrapper.


Oddant1 (approved):
Y'know this looks an awful lot like that other thing I said was good, so this is probably good too! As good as R ever is anyway.

PR #338, opened 2023-01-10
ENH: Add partial Procrustes support

gibsramen:
Co-authored-by: @wasade

Adds support for a modified Procrustes analysis wherein a computed transformation can be fit on two ordinations and then applied to additional points. We envision this being used, for example, to compare paired samples sequenced with WGS and 16S.

On the Q2 interface, we are thinking of adding a parameter to `qiime diversity procrustes-analysis` that can disable strictness on ID overlap. Additional things to consider are modifying the visualization to show (a) M^2 by increasing number of paired samples used and (b) before/after Pearson correlation of PC1/PC2/PC3 for the paired samples.

Daniel has used the code added in the initial commit to some encouraging preliminary success. Below is an example of a subset of 200 paired samples (out of 800 total) upon which both WGS and 16S were performed.

image

image

### DIFF
--- a/q2_diversity/_procrustes.py
+++ b/q2_diversity/_procrustes.py
@@ -117,11 +117,9 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
     return df
 
 
-def deconstructed_procrustes(mtx1: np.array,
-                             mtx2: np.array) -> (np.ndarray, np.ndarray,
-                                                 float, float, np.ndarray,
-                                                 float):
-    """Derived from scipy procrustes."""
+def _deconstructed_procrustes(mtx1, mtx2):
+    # Derived from scipy procrustes
+    # https://github.com/scipy/scipy/blob/d541c752246a9e196034957d3e044950eec75907/scipy/spatial/_procrustes.py#L100-L125
     mtx1 = mtx1.copy()
     mtx2 = mtx2.copy()
 
@@ -147,10 +145,7 @@ def deconstructed_procrustes(mtx1: np.array,
     return mtx1_translate, mtx2_translate, norm1, norm2, R, s
 
 
-def partial_procrustes(df_mtx1: pd.DataFrame,
-                       df_mtx2: pd.DataFrame,
-                       df_mtx1_pair_ids: list,
-                       df_mtx2_pair_ids: list) -> (pd.DataFrame, pd.DataFrame):
+def _partial_procrustes(df_mtx1, df_mtx2, df_mtx1_pair_ids, df_mtx2_pair_ids):
     df_mtx1 = df_mtx1.copy()
     df_mtx2 = df_mtx2.copy()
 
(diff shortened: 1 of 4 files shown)


ebolyen:
Hey @gibsramen and @wasade, this looks really cool and we agree it would be very helpful for multi-omics.

The proposal of adding a[/some] flags to the existing action also sounds good. Let us know if you need anything!

q2_diversity/_procrustes.py
```diff
@@ -114,3 +117,101 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
...
+        raise ValueError('Cannot fit fewer dimensions than available')
+
+    pairing = pairing.to_series()
+    pairing = pairing[~pairing.isnull()]
+
+    if len(pairing) == 0:
+        raise ValueError('The metadata are lacking paired samples')
+
+    ref_pairs = sorted(set(pairing.index) & set(reference.samples.index))
+    other_pairs = sorted(set(pairing.index) & set(other.samples.index))
```
wasade, line 138: It is not assured that the set of samples represented here are the same, if for instance, one ordination lacked a sample. 

```suggestion
    pairing = pairing.to_series()
    pairing = pairing[~pairing.isnull()]
    pairing = pairing.loc[set(reference.samples.index) & set(other.samples.index)]

    if len(pairing) == 0:
        raise ValueError('The metadata are lacking paired samples')

    ref_pairs = sorted(set(pairing.index) & set(reference.samples.index))
    other_pairs = sorted(set(pairing.index) & set(other.samples.index))
```

q2_diversity/_procrustes.py
```diff
@@ -114,3 +117,101 @@ def _procrustes_monte_carlo(reference: np.ndarray, other: np.ndarray,
...
+
+    if other.samples.shape[1] < dimensions:
+        raise ValueError('Cannot fit fewer dimensions than available')
+
+    pairing = pairing.to_series()
+    pairing = pairing[~pairing.isnull()]
+
+    if len(pairing) == 0:
+        raise ValueError('The metadata are lacking paired samples')
+
```
wasade, line 136: Rob raise that we need at least +1 point more than the number of dimensions

wasade: ...do check from whats in reference/other. if missing do diagnostics at end

wasade: Done

wasade:
@ebolyen, do you think this is feasible for merge before code freeze?

wasade:
@ebolyen just checking on this, do you know when it may be feasible for this to be merged?

ebolyen (approved):
I'm really sorry I lost track of these pings @wasade. 

This code is quite lovely and while I can't speak to the math or tests directly, the numpy looks good.

wasade:
Thank you, @ebolyen!!

PR #343, opened 2023-08-14
DEP: updating numba to solve environment for unifrac updates

lizgehret:


ebolyen:
Related PR: https://github.com/qiime2/distributions/pull/4

PR #348, opened 2024-01-04
BUG: Exposes 'ignore_missing_samples' parameter in core metrics from emperor

### LINKED ISSUE #347: BUG: `ignore_missing_samples` param not exposed from emperor plot in `core-metrics` and `core-metrics-phylogenetic`
**Bug Description**
When running `core-metrics` (and `core-metrics-phylogenetic`), the following error will be raised if a user's metadata file is missing sample IDs that are present in the feature table:

Screen Shot 2023-12-05 at 1 37 20 PM

[This error is raised from within emperor](https://github.com/biocore/emperor/blob/fcb4735c73d3e0f24c4127c9d5e07ca5ee020e58/emperor/core.py#L322C1-L322C75), and the suggestion to enable the `ignore_missing_samples` parameter is not possible, because this parameter is not passed from either of the pipelines above into the emperor plot action.

**Steps to reproduce the behavior**
1. Go to the [Moving Pictures tutorial](https://docs.qiime2.org/2023.9/tutorials/moving-pictures/) and download the following files:
`sample-metadata.tsv`
`table-dada2.qza`

2. Open `sample-metadata.tsv` and delete a few rows of data.

3. In a QIIME 2 2023.9 Amplicon environment (just for consistency's sake - this should produce the same result regardless of the specific version), run the following command:
```
qiime diversity core-metrics \
--i-table table-dada2.qza \
--p-sampling-depth 1103 \
--m-metadata-file sample-metadata.tsv \
--output-dir core-metrics-outputs \
--verbose
```
The same error as in the above screenshot will be produced.

**Expected behavior**
The `ignore_missing_samples` parameter should be added as an input parameter to the `core-metrics` and `core-metrics-phylogenetic` pipelines so that it can be passed into the emperor plot action.

hagenjp:


hagenjp:
@lizgehret just waiting on CI but this is ready for your review

q2_diversity/__init__.py
```diff
@@ -27,5 +27,5 @@
            'core_metrics_phylogenetic', 'core_metrics',
            'filter_distance_matrix', 'mantel', 'alpha_rarefaction',
            'beta_rarefaction', 'procrustes_analysis', 'beta_correlation',
-           'adonis', 'partial_procrustes'
+           'adonis', 'partial_procrustes', 'ignore_missing_samples'
```
lizgehret:
```suggestion
           'adonis', 'partial_procrustes'
```

lizgehret: This isn't needed here since we aren't adding a new method (just exposing a parameter on an existing method).

lizgehret (changes requested):
generally looks good @hagenjp, just one small comment inline! i'm pulling down to test locally right now, and will follow up if anything comes up from there.

lizgehret:
Hey @hagenjp - couple of things came up while I was testing. Quick overview below, but let's meet on Monday to work on this together!

1. When running core-metrics using the steps listed in the original issue, the following error is raised:
Screen Shot 2024-01-05 at 5 19 26 PM

2. When running core-metrics-phylogenetic, the `ignore_missing_samples` parameter is not recognized.

Offhand, I'm not sure why 1. is happening - but I know why 2. is happening. These will be good things to pair on, and this is a great example of why we always test locally before merging - all of the existing tests pass, but this functionality isn't working like we want it to 🙃

lizgehret:
Okay so for posterity - @hagenjp and I met to discuss the `index cannot be a set` error shown in the screenshot above.

The issue seems to be coming from the `_validate_metadata` helper step within `emperor`, specifically when `index` is assigned to the `difference` variable, which is assumed to be a set. `pandas >=1.5` does not allow sets to be used in indices or columns, since they are not 'array-like' - which was a breaking change.

We tested that the error was related to this change in pandas 1.5 by creating a QIIME 2 core 2022.8 environment (our latest release prior to upgrading to pandas 1.5). One side note is that we had to set the conda `--solver` flag to `classic` in order to utilize our 2022.8 environment file (the libmamba solver couldn't solve the environment). That's a separate issue to discuss, but adding here for the record.

Once in a 2022.8 environment, we created a local branch for q2-diversity based off of the 2022.8 release commit and then cherry-picked @hagenjp's commits from this PR to ensure that we were working in a 2022.8 version of q2-diversity that also included the changes from this PR. We then re-ran `core-metrics` with the steps described in the associated issue, which were successful.

Next steps are to determine if it's possible to get around this issue in q2-emperor, or if we need to hold off on this PR until the necessary changes are made in emperor (if this is the case, we'll open an issue there).

@hagenjp I'd recommend reviewing everything we did above, and seeing if you can replicate all of this on your end without me. This is a great learning experience with traceback reading, conda, and git 🙂

lizgehret:
Okay so final update on this:

We cannot workaround this in q2-emperor, and the culprit lies [here](https://github.com/biocore/emperor/blob/fcb4735c73d3e0f24c4127c9d5e07ca5ee020e58/emperor/core.py#L301) within emperor. However, a fix has been rolled out [here](https://github.com/biocore/emperor/commit/6163b4e6f549983ba0a7be9975e765ce55747c95) so that the `difference` object is converted to a list before being passed into the `index` param. These changes won't be available until the next version of emperor is released, which looks like (from [this commit](https://github.com/biocore/emperor/commit/fcb4735c73d3e0f24c4127c9d5e07ca5ee020e58)) is slated for July 2024. So we'll have to wait until we can upgrade to this version of emperor before this PR will pass.

We'll leave this open and marked as blocked until we upgrade to the next version of emperor.

ElDeveloper:
@lizgehret thanks for tracking this down. I just noticed I never pushed a release to GitHub when I published 1.0.4 to PyPI (oops!). This is fixed, and I've pushed a pull request to q2-emperor:

https://github.com/qiime2/q2-emperor/pull/108

lizgehret:
> @lizgehret thanks for tracking this down. I just noticed I never pushed a release to GitHub when I published 1.0.4 to PyPI (oops!). This is fixed, and I've pushed a pull request to q2-emperor:
> 
> [qiime2/q2-emperor#108](https://github.com/qiime2/q2-emperor/pull/108)

Thanks @ElDeveloper! We'll hold off merging that PR until emperor 1.0.4 is available on conda - you're expecting that to be sometime in July of this year?

ElDeveloper:
@lizgehret Better than that, 1.0.4 has been available via conda-forge for the past 6 months:
https://anaconda.org/conda-forge/emperor/files

🎉 🐧

lizgehret:
@ElDeveloper oh wow, that didn't come up when I searched for available emperor builds on anaconda - in any case, this is great! Thanks! I'll get this PR merged 🙂

lizgehret:
**EDIT**: Okay so additional follow up re: emperor 1.0.4 - I created a 2023.9 amplicon environment with @ElDeveloper's PR branch installed locally (that updates emperor within q2-emperor's recipe to 1.0.4). Conda can solve the environment, so this should be compatible with the amplicon distro. q2-emperor's tests all pass locally w/this version, ~~but we have 3 test failures in q2-diversity to examine (the only other plugin w/an indirect dependency on emperor via q2-emperor)~~ and all tests pass in q2-diversity as well. The functionality in this PR works as expected with the updated version of emperor. ~~So once the test failures in q2-diversity are sorted,~~ We should be good to merge the version pin update PR as well as this one!

lizgehret:
Okay @hagenjp, I've gone ahead and merged the version pin update PR in q2-emperor. So remaining to-do's here are:

- [x] When running core-metrics-phylogenetic, the `ignore_missing_samples` parameter is not recognized. This parameter is missing somewhere within the core-metrics-phylogenetic pipeline - take a look in there and add that where it's needed 🙂 
- [x] Add tests for this new exposed parameter - we'll want to assert that we get a failure in core-metrics and core-metrics-phylogenetic for the original use-case, and that enabling this flag will resolve that failure in each case.

Make sure to pull down and install the latest version of q2-emperor with the updated pin in your development environment, and `conda install -c conda-forge emperor=1.0.4` in that same environment. You should then be able to effectively test all of this locally! Lmk if you have any questions 🙂

hagenjp:
@lizgehret If this passes after the build on Sunday it will be ready for your review! Thanks :)

q2_diversity/_core_metrics.py
```diff
@@ -57,7 +59,8 @@ def core_metrics_phylogenetic(ctx, table, phylogeny, sampling_depth, metadata,
 
     cr = core_metrics(table=table, sampling_depth=sampling_depth,
                       metadata=metadata, with_replacement=with_replacement,
-                      n_jobs=n_jobs_or_threads)
+                      n_jobs=n_jobs_or_threads,
+                      ignore_missing_samples=ignore_missing_samples)
```
lizgehret, line 63: Nice! Good find on this one 🙂 this is a great example of how pipelines can depend on other pipelines. Core metrics phylogenetic is just regular core metrics being called plus a few extra methods (which are the ones that use phylogeny as an input).

q2_diversity/tests/test_core_metrics.py
```diff
@@ -144,6 +144,84 @@ def test_core_metrics(self):
...
+        table = biom.Table(np.array([[150, 100, 100], [50, 100, 100]]),
+                           ['O1', 'O2'],
+                           ['S1', 'S2', 'S3'])
+        table = Artifact.import_data('FeatureTable[Frequency]', table)
+
+        metadata = Metadata(
+            pd.DataFrame({'foo': ['1', '2']},
+                         index=pd.Index(['S1', 'S2'], name='id')))
+
+        with self.assertRaisesRegex(KeyError, 'samples not included'):
```
lizgehret: This looks good, but I would also include the particular sample ID that isn't included in the metadata in your `assertRaisesRegex` call (that way we're being quite specific on the error being raised including the correct missing samples).

q2_diversity/tests/test_core_metrics.py
```diff
@@ -144,6 +144,84 @@ def test_core_metrics(self):
...
+
+        tree = skbio.TreeNode.read(io.StringIO(
+            '((O1:0.25, O2:0.50):0.25, O3:0.75)root;'))
+        tree = Artifact.import_data('Phylogeny[Rooted]', tree)
+
+        metadata = Metadata(
+            pd.DataFrame({'foo': ['1', '2']},
+                         index=pd.Index(['S1', 'S2'], name='id')))
+
+        with self.assertRaisesRegex(KeyError, 'samples not included'):
```
lizgehret: Same comment as w/your core metrics test (w/ignore missing samples set to false)

lizgehret (commented):
Nice job @hagenjp! Couple of small notes on your unit tests, but otherwise this should be ready to merge next week once we confirm ci passes with the new q2-emperor build.

lizgehret (approved):
great work @hagenjp! this one is ready to merge 🙂

PR #350, opened 2024-02-12
MAINT: add threads/jobs primitives

colinvwood:


lizgehret (approved):

lizgehret:
@colinvwood and I spoke offline about this - the reason this is failing is because diversity directly depends on diversity-lib, and the changes implemented in that repo were not included in the latest prepare (since it was just merged along with the rest of the threads/jobs PRs). these failures essentially aren't real - so i'm going to merge this, and we'll use the next prepare (that we need to run anyways to add rescript) to confirm that everything behaves as expected.

PR #351, opened 2024-02-20
BUG: improves error message to describe too few samples retained in '…

### LINKED ISSUE #346: BUG: more descriptive error message for too few samples retained in 'alpha-group-significance' visualization
**Bug Description**:
When running `alpha-group-significance`, this is one of the possible errors that can be produced:

Screen Shot 2023-12-01 at 3 43 08 PM

The language in this message implies that the error stems from the contents of the metadata file (i.e. doesn't contain columns that meet requirements such as containing categorical data, not being empty, etc). However, a forum user reached out with this issue ([forum post link](https://forum.qiime2.org/t/alpha-group-significance-error-message/28307)) and upon examination of their data, the source of the error was due to the fact that they had chosen a sampling depth that only retained 2 samples. This, in turn, resulted in the metadata file pulling information on those two samples that didn't meet the stated requirements of the visualizer - specifically, that values in the column were seen as unique, since there were only two samples and the column in question had groupings that were different for the two samples.

This error message should include details regarding the sample IDs that are being looked up in the metadata file, and the language should be changed to include the possibility that one's chosen sampling depth may be too deep, causing too few samples to be retained for the visualizer.

**Steps to reproduce the behavior**:
1. Download the files shared from the OP in the linked forum post above.
2. In a 2023.9 QIIME 2 Amplicon environment, run the following command (in the same dir as the above files):
```
qiime diversity alpha-group-significance \
--i-alpha-diversity shannon_vector.qza \
--m-metadata-file Testmetadata.tsv \
--o-visualization shannon_vector.qzv \
--verbose
```
_(Error message from above will be produced)_

3. Unzip the `shannon_vector.qza` artifact and examine the alpha-diversity.tsv file contained within the data directory. The following two IDs will be the only ones present: 98039, W190020
4. Open `Testmetadata.tsv` and examine the contents of the metadata associated with the two IDs above. These do not meet the stated requirements of the visualizer because within the `Location` column, the value associated with sample 98039 is `zoo` and the value associated with sample W190020 is `wild`.
 
These don't satisfy the following requirement:
```
There must be at least one metadata column doesn't consist of unique values.
```

**Expected behavior**:
The error message should include something like 'the contents of the metadata file _associated with the samples present_ in the alpha-diversity metric' and should include the sample IDs in the error message. It is probably worth re-writing the message to clearly state that this error can result of either a metadata file that doesn't meet the viz requirements, or samples whose associated metadata doesn't meet the requirements (with a suggestion to check the diversity metric to ensure an appropriate sampling depth was selected).

hagenjp:


lizgehret:
just FYI @hagenjp these failures aren't on you - currently dev bumping CI for the new epoch, and that's why things are failing. should be fixed in a couple hours and then we can re-run these jobs!

q2_diversity/tests/test_alpha.py
```diff
@@ -439,6 +439,22 @@ def test_alpha_group_significance_forward_slash_in_metadata_col(self):
...
+    def test_sample_ids_in_error_message(self):
+        alpha_div = pd.Series([2.0, 4.0, 6.0], name='alpha-div',
+                              index=['sample1', 'sample2', 'sample3'])
+        md = qiime2.Metadata(
+            pd.DataFrame(
+                {'col1': [1, 2, 1],
+                 'col2': [4.2, 4.2, 4.3]},
+                index=pd.Index(['sample1', 'sample2', 'sample3'], name='id')))
+
+        with self.assertRaises(ValueError) as cm:
```
lizgehret: Mainly for my own edification, but what was your reasoning behind choosing `cm` as the variable name here?

q2_diversity/_alpha/_visualizer.py
```diff
@@ -48,12 +48,20 @@ def alpha_group_significance(output_dir: str, alpha_diversity: pd.Series,
...
 
     if len(metadata.columns) == 0:
+        sample_ids = ', '.join(alpha_diversity.index)
         raise ValueError(
-            "Metadata does not contain any columns that satisfy this "
-            "visualizer's requirements. There must be at least one metadata "
-            "column that contains categorical data, isn't empty, doesn't "
-            "consist of unique values, and doesn't consist of exactly one "
-            "value.")
+            "Either the metadata file does not meet the requirements of this "
```
lizgehret, line 53: Generally, this lgtm! I'll pull this down locally to test and make sure nothing looks weird on the command line, but otherwise this should be g2g 👍

lizgehret (commented):
thanks @hagenjp! i just need to test locally and then this should be good to merge.

q2_diversity/tests/test_alpha.py
lizgehret: hey @hagenjp i went ahead and updated this variable name to `ve` (which is commonly what we do for ValueError abbreviations)

lizgehret:
local testing looks good, thanks @hagenjp!

PR #358, opened 2024-05-23
BUG: fixed swapped labels for beta_correlation action

cherman2:
Fixes labels for beta_correlations so that it correctly labels the metadata axis and the distance matrix axis.


gregcaporaso:
Tested locally and it looks good, thanks @cherman2!

PR #360, opened 2024-06-11
BUG: x-axis data missing in alpha rarefaction viz

### LINKED ISSUE #359: BUG: alpha_rarefaction viz has lost sequencing depth data
In 2024.5, the alpha rarefaction plot no longer contains sequencing depth data (i.e. x-axis data), so all of the diversity metric data is stacked at zero on the y-axis (see images below):
Screen Shot 2024-06-11 at 1 00 22 PM

Screen Shot 2024-06-11 at 1 05 44 PM

Original report was from this forum [xref](https://forum.qiime2.org/t/https-forum-qiime2-org-t-alpha-rarefaction-gzv-is-not-displayed-9012/30432), but I confirmed this is happening with the alpha rarefaction plots in [moving pictures](https://docs.qiime2.org/2024.5/tutorials/moving-pictures/#alpha-rarefaction-plotting) & the [cancer microbiome](https://docs.qiime2.org/jupyterbooks/cancer-microbiome-intervention-tutorial/030-tutorial-downstream/040-even-sampling.html#alpha-rarefaction-plots) tutorials as well.

No recent changes have occurred within the visualizer functionality outside of the python 3.9 updates ([PR link](https://github.com/qiime2/q2-diversity/pull/356) for reference), which don't seem to be related from my initial investigation.

lizgehret:
fixes #359 

So this does work, but I'm not happy with the fact that this had to be added (seemingly) randomly. I'm signing off for the day, but will try to find a better solution that addresses what changed from 2024.2 to 2024.5 tomorrow.

### DIFF
--- a/q2_diversity/_alpha/_visualizer.py
+++ b/q2_diversity/_alpha/_visualizer.py
@@ -273,10 +273,6 @@ def _compute_summary(data, id_label, counts=None):
         summary_df['count'] = 1
     summary_df = summary_df.reset_index()
     summary_df.rename(columns={'level_0': id_label}, inplace=True)
-    # this seems to work, but it's unclear to me why this wasn't needed
-    # prior to 2024.5 release & python 3.9 update
-    summary_df.rename(columns={'level_1': '_alpha_rarefaction_depth_column_'},
-                      inplace=True)
     return summary_df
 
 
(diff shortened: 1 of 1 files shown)


Oddant1:
@lizgehret the image in this viz is in svg, so each element of the viz has actual DOM nodes. That means in theory we can test it with selenium. If we are patching, we shouldn't hold up the patch for that, but it's probably worth adding those tests. If only they had been added ages ago lol.

lizgehret:
> @lizgehret the image in this viz is in svg, so each element of the viz has actual DOM nodes. That means in theory we can test it with selenium. If we are patching, we shouldn't hold up the patch for that, but it's probably worth adding those tests. If only they had been added ages ago lol.

I agree on both counts - we should add tests, but we should wait until after the patch to do so. Want to walk me through selenium testing next week when I'm back in office?

Oddant1:
Sure

lizgehret:
**update 1:** viz works as expected without metadata as input.

**update 2**: [this](https://github.com/qiime2/q2-diversity/blob/5fe9c04e39ca4c6028653e0378bf18b079930d9b/q2_diversity/_alpha/_visualizer.py#L398) is the culprit. the updated fix ensures that these names match the multi-index DF, so that the names aren't overwritten in the case that they don't match up when the merge occurs.

PR #276, opened 2020-06-09
Bioenv 0 variance filter

Oddant1:
closes #275

Issue #275, opened 2020-06-09
Bioenv fails if 0 variance columns are left after dropping rows with empty cells


Oddant1:
**Bug Description**
If `bioenv` gets any 0 variance numerical columns, it errors. There is currently a safeguard in place to prevent 0 variance columns from ever reaching the stage where they become an issue; however, after this safeguard we drop all rows containing any empty cells. If any column becomes 0 variance after dropping these rows, we still get the error.

**Steps to reproduce the behavior**
1. Run the following commands
```
wget \
  -O "sample-metadata.tsv" \
  "https://data.qiime2.org/2020.2/tutorials/atacama-soils/sample_metadata.tsv"

qiime metadata distance-matrix \
  --m-metadata-file sample-metadata.tsv \
  --m-metadata-column elevation \
  --o-distance-matrix elevation-distance-matrix.qza

qiime diversity bioenv \
  --i-distance-matrix elevation-distance-matrix.qza \
  --m-metadata-file sample-metadata.tsv \
  --o-visualization bioenv-elevation.qzv \
  --verbose
```
2. Error

**Expected behavior**
Filter zero variance columns after removal of rows containing empty cells so this doesn't happen.

**Screenshots**
![image](https://user-images.githubusercontent.com/10642616/84188191-09fa7e80-aa48-11ea-9dae-6aab892bcc8a.png)

**Computation Environment**
- OS: Linux Mint 19.3 (forum user was using virtual box)
- QIIME 2 2020.2

**References**
1. [Forum](https://forum.qiime2.org/t/qiime-diversity-bioenv-error-when-running-through-atacama-soil-microbiome-tutorial/15272/7) post that prompted this.
2. Ancient [forum](https://forum.qiime2.org/t/diversity-bioenv-column-s-in-the-data-frame-could-not-be-scaled/58/3) post in regards to the same thing.
3. Ancient [PR](https://github.com/qiime2/q2-diversity/pull/72) discussed in that forum post that seems to have been superceded.

ChrisKeefe:
Again on the [forum](https://forum.qiime2.org/t/bioenv-error-atacama-tutorial/15399/2)

PR #361, opened 2024-07-09
BUG: fix bioenv erroring on MD columns without any data

### LINKED ISSUE #320: Bioenv raises a ValueError when a metadata file has columns with no values, even when other columns have values
**Bug Description**
If a metadata file has one or more columns with no values in any cells, bioenv will panic, even if there are several other columns with data in every cell. This behavior was not present in qiime version 2020.2, and was likely introduced by https://github.com/qiime2/q2-diversity/commit/76e368fb21d1a9d5a950e4387011680b0fac2859. Originally reported by me on the forums.

**Steps to reproduce the behavior**
1. Use these files: [bioenv_files.tar.gz](https://github.com/qiime2/q2-diversity/files/7555979/bioenv_files.tar.gz)
2. Run the command `qiime diversity core-metrics-phylogenetic --verbose --i-phylogeny rooted-tree.qza --i-table table-trim55126.qza --m-metadata-file 203967.metadata.tsv --p-sampling-depth 300 --output-dir ./output/`
3. Run the command `qiime diversity bioenv --verbose --i-distance-matrix ./output/unweighted_unifrac_distance_matrix.qza --m-metadata-file 203967.metadata.tsv --o-visualization ./output/unweighted-unifrac-bioenv.qzv`

**Expected behavior**
I should see the output:
`Saved Visualization to: ./output/unweighted-unifrac-bioenv.qzv`

**Computation Environment**
- OS: Rocky Linux 8.4 with Conda
- QIIME 2 Release: 2021.4 and 2021.8 tested

**References**
1. https://forum.qiime2.org/t/bioenv-diversity-error-with-qiime-version-2021-4-but-not-with-2020-2/21319

lizgehret:
Fixes: #320 

For context, this was originally introduced by issue #275 (linked PR #276). I'm a bit surprised that the initial fix didn't involve dropping the empty columns (vs. dropping the rows with any empty values).

From the OP's example data, here's the resultant viz after dropping empty columns from their data:
Screen Shot 2024-07-09 at 4 50 48 PM

I'll need to take a closer look at what the expected output should be to see if this is actually a reasonable fix.

### DIFF
--- a/q2_diversity/_beta/bioenv_assets/index.html
+++ b/q2_diversity/_beta/bioenv_assets/index.html
@@ -3,26 +3,25 @@
 {% block content %}
     {% if non_numeric_cols %}
       <div class="alert alert-warning col-md-12">
-        The following metadata columns have been omitted because the metadata
-        column was not numeric:
+        The following metadata column(s) have been omitted
+        because they were not numeric:
         <strong>{{ non_numeric_cols }}</strong>
       </div>
     {% endif %}
     {% if zero_variance_cols %}
       <div class="alert alert-warning col-md-12">
-        The following metadata columns have been omitted because the metadata
-        column had no variance, or consisted only of missing values:
+        The following metadata column(s) have been omitted
+        because they had no variance:
         <strong>{{ zero_variance_cols }}</strong>
       </div>
     {% endif %}
-    {% if initial_dm_length != filtered_dm_length %}
-      <p class="alert alert-warning">
-        <b>Warning</b>: Some samples were filtered from the input distance matrix
-        because they were missing metadata values.<br><b>The input distance matrix
-        contained {{ initial_dm_length }} samples but bioenv was computed on only
-        {{ filtered_dm_length }} samples.
-      </p>
-    {% endif %}
+    {% if missing_values_cols %}
+    <div class="alert alert-warning col-md-12">
+      The following metadata column(s) have been omitted
+      because they contained missing values:
+      <strong>{{ missing_values_cols }}</strong>
+    </div>
+  {% endif %}
     <div class="row">
         <div class="col-lg-12">
             {{ result }}
--- a/q2_diversity/tests/test_beta.py
+++ b/q2_diversity/tests/test_beta.py
@@ -385,88 +385,81 @@ class BetaDiversityTests(TestPluginBase):
 class BioenvTests(TestPluginBase):
     package = 'q2_diversity.tests'
 
-    def test_bioenv(self):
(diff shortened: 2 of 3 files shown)


lizgehret:
okay i'm happier with the tests now - i've also updated the filtering functionality to filter out sparse columns (vs. filtering out samples).

q2_diversity/tests/test_beta.py
```diff
@@ -385,88 +385,81 @@ def test_beta_phylogenetic_too_many_jobs(self):
...
             index_fp = os.path.join(output_dir, 'index.html')
+
+            # assert that the html file exists
             self.assertTrue(os.path.exists(index_fp))
-            self.assertTrue('metadata1' in open(index_fp).read())
 
-            self.assertTrue('not numeric:' in open(index_fp).read())
-            self.assertTrue('<strong>metadata2' in open(index_fp).read())
+            # assert that metadata1 was found in the table
+            self.assertTrue('<th>metadata1</th>' in open(index_fp).read())
```
colinvwood: There are a lot of tests that follow this pattern. Using `assertIn` is more readable, and just performing `open(index_fp).read()` once and storing it would be better. (I know that both of these issues already existed but might as well fix them while we're here.)

q2_diversity/_beta/bioenv_assets/index.html
```diff
@@ -3,26 +3,25 @@
...
+        because they had no variance:
         <strong>{{ zero_variance_cols }}</strong>
       </div>
     {% endif %}
-    {% if initial_dm_length != filtered_dm_length %}
-      <p class="alert alert-warning">
-        <b>Warning</b>: Some samples were filtered from the input distance matrix
-        because they were missing metadata values.<br><b>The input distance matrix
-        contained {{ initial_dm_length }} samples but bioenv was computed on only
-        {{ filtered_dm_length }} samples.
```
colinvwood, line 23: Did you figure that this just isn't really that useful to know?

lizgehret: @colinvwood I modified the approach for what's retained in this calculation - instead of removing sparse rows (i.e. samples), we're now removing sparse columns. This ensures maximal retention of samples and instead removes sparse measures, vs. sacrificing samples associated with any sparse columns (which can render the entire dataframe empty if there are multiple sparse columns that span the entire index length).

q2_diversity/tests/test_beta.py
lizgehret: makes sense, i'll update the tests accordingly 👍🏻

colinvwood (approved):

PR #379, opened 2025-06-07
Single sample issue

### LINKED ISSUE #241: Validating the number of samples
**Improvement Description**
A user on the [QIIME 2 forum](https://forum.qiime2.org/t/core-metrics-phylogenetic-crashed-free-invalid-next-size/8408) encountered a memory error when running a `FeatureTable` containing a single sample through `core-metrics-phylogenetic`. The error stemmed from Striped UniFrac not appropriately checking for a single sample (noted in biocore/unifrac#34). While the actual error stemmed from underlying code, we may want to consider adding in a helpful message to the user that this action is not relevant for a single sample.

**Current Behavior**
When a `FeatureTable` with a single sample is provided, and error is produced.

**Proposed Behavior**
Error, and provide a polite message to the user indicating that `beta` metrics are only applicable to tables with 2 or more samples.

Macabe222:
Produces an error when a user passes in a feature table with less than two samples. Explains that later in the  `core_metrics_phylogenetic` pipeline the feature table is used for beta diversity. 

Adresses qiime2/q2-diversity#241.


colinvwood (approved):

q2_diversity/_core_metrics.py
```diff
@@ -48,6 +49,13 @@ def core_metrics(ctx, table, sampling_depth, metadata, with_replacement=False,
 def core_metrics_phylogenetic(ctx, table, phylogeny, sampling_depth, metadata,
                               with_replacement=False, n_jobs_or_threads=1,
                               ignore_missing_samples=False):
+    biom_table = table.view(biom.Table)
+    if biom_table.length() < 2:
+        raise ValueError(
+            'Table must have at least two samples as beta diversity will be'
+            ' applied later.'
+        )
```
colinvwood: Realized that this check will be better off in the `core_metrics` pipeline because it's a useful check there too, and that pipeline is called by this one so both will be covered

PR #391, opened 2026-02-18
LANG: test fix for unifrac & friends updates (amplicon)

lizgehret:


lizgehret:
ci failure isn't interesting - but just for posterity, this is why i'm doing the unifrac & friends dependency updates. we've successfully removed `pkg_resources` from being brought into our environments, which now causes a failure bc the older version of unifrac is calling it. this update bumps unifrac to the newer version which includes my `pkg_resources` deprecation changes, which will fix this failure. in the process of updating unifrac, several other deps needed updates to solve the environment, which in turn, introduced a couple of breaking API changes - this one in particular is a change in numpy (from np.Infinity -> np.inf).

lizgehret:
we should hold off merging this until all other test failures are addressed in the parent [LANG PR](https://github.com/qiime2/distributions/pull/1162).

ebolyen (approved):

PR #392, opened 2026-03-06
LANG: updates for 2026.4

lizgehret:
supercedes #385 

This PR includes the following changes:
- minor param name changes from numpy updates
- adonis -> adonis2 implementation & test improvements
- nj test improvements to accomodate skbio updates

PR #393, opened 2026-03-25
BUG: fix missing metadata handling after pandas update

ebolyen:
A tiny bit of fallout from the update in #392 (the `(c, '')` moved to `(c, '_')` which is necessary as pandas seems to simplify the former oddly. In fact it seems to have lived in a very ambiguous place, so the empty string was ignored by some methods.

Now that it gets standardized more uniformly, we need to use a real empty placeholder like `_`, but this has a few downstream impacts which weren't caught until we had missing metadata.

### DIFF
--- a/q2_diversity/_alpha/_visualizer.py
+++ b/q2_diversity/_alpha/_visualizer.py
@@ -242,7 +242,11 @@ def alpha_correlation(output_dir: str,
 
 
 def _reindex_with_metadata(column, columns, merged):
-    reindexed = merged.set_index(column)
+    # Metadata columns are padded to a 2-level MultiIndex to align with the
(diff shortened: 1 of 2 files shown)


ebolyen:
Just ran gut-to-soil and the plot looks good. (extension renamed for github, [q2view will accept .zip btw])
[alpha-rarefaction-dcb5fb20b92db5bf15ab099076cae6cb.zip](https://github.com/user-attachments/files/26240989/alpha-rarefaction-dcb5fb20b92db5bf15ab099076cae6cb.zip)

lizgehret (approved):
all makes sense, thanks @ebolyen!

